PR #59, opened 2017-12-21
changes to contract view and edit mode

Eelsie:


src/leases/components/leaseSections/contract/Contract.js
```diff
@@ -27,40 +27,75 @@ const Contract = ({contract}: Props) => {
...
               <label>Allekirjoituspäivämäärä</label>
               <p>{contract.signing_date ? helpers.formatDate(get(contract, 'signing_date', '–')) : '–'}</p>
-
-              <label>Vuokravakuusnumero</label>
-              <p>{contract.lease_deposit_number ? get(contract, 'lease_deposit_number', '–') : '–'}</p>
-
+            </Column>
+            <Column medium={4}>
+              <label>Kommentti allekirjoitukselle</label>
+              <p>{contract.signing_date_comment ? get(contract, 'signing_date_comment', '–') : '–'}</p>
```
jorilindell: Default string (-) is defined twice. You get same end result with
<p>{get(contract, 'signing_date_comment', '–')}</p>

Same thing on rows 43, 59, 73, 87, 91, 95

src/leases/components/leaseSections/contract/Contract.js
```diff
@@ -27,40 +27,75 @@ const Contract = ({contract}: Props) => {
...
+            </Column>
+          </Row>
+          {contract.pledge_books.length > 0 &&
+          <Row>
+            <Column>
+              <p className='green-subtitle'>Panttikirjat</p>
+            </Column>
+          </Row>
+          }
+          {contract.pledge_books.length > 0 && contract.pledge_books.map((pledge_book, index) =>
```
jorilindell: Page will crash if contract.pledge_books is null. I would be good to check is contract.pledge_books defined
e.g.
{contract.pledge_books && contract.pledge_books.length > 0 && contract.pledge_books.map((pledge_book, index) =>

src/leases/components/leaseSections/contract/Contract.js
Eelsie: The purpose is to show '–' also if null

src/leases/components/leaseSections/contract/Contract.js
Eelsie: of course! forgot that :/

PR #256, opened 2019-08-21
Delete invoice

kovedev:


kovedev:
Delete invoice and display detail in error modal.

PR #346, opened 2021-09-27
Basic plot search form template functionality

tomimakela:
Allows the selection of a template form, which will be duplicated as its own form instance when the plot search is saved. Also allows replacing that saved form with a new template and previewing the fields included in the current form.


tomimakela:
Was developed against backend commit `23c60ddc`, apparently doesn't work quite right with the up-to-date backend anymore.

PasiVuohijoki (approved):
Great job!

PR #374, opened 2022-04-01
Feature/applications

PasiVuohijoki:
Plotapplication listing and searchin/filtering is working. Template for single application page made.


src/plotApplications/components/map/SingleLeaseMap.js
```diff
@@ -0,0 +1,297 @@
+// @flow
```
tomimakela, line 1: Um. I read through the files for AreasLayer, PlanUnitsLayer, PlotsLayer and SingleLeaseMap and even made some comments on them, but on closer inspection it seems they're just straight copies of files from another view and aren't used anywhere...

src/plotApplications/components/map/TargetListLayer.js
```diff
@@ -0,0 +1,78 @@
+// @ flow
```
tomimakela, line 1: The majority of this file is the same code as the other layer components (from the leases view). I guess we can do it this way since there's already a precedent, even if I'd really want to write an auxiliary component in between that only takes the popup content and handles everything else by itself.

src/plotApplications/components/map/TargetListLayer.js
```diff
@@ -0,0 +1,78 @@
...
+      data={targetsGeoJson}
+      onEachFeature={(feature, layer) => {
+        if (feature.properties) {
+          const {id, identifier} = feature.properties;
+
+          const getApplicationsLink = () => {
+            return `${getRouteById(Routes.PLOT_APPLICATIONS)}/?visualization=table&identifier=${identifier}`;
+          };
+
+          const popupContent = `<p><strong>Id:</strong> <a href=${getApplicationsLink()}>${identifier}</a></p>`;
```
tomimakela, line 50: This is rather bare; it should be very possible to add the target address for example in getApplicationTargetFeatures and then use it here.

src/plotApplications/components/PlotApplicationsListPage.js
```diff
@@ -265,48 +442,46 @@ class PlotApplicationsListPage extends PureComponent<Props, State> {
...
+              onSearch={this.handleSearchUpdated}
+              states={selectedStates}
             />
           </Column>
         </Row>
-
         <TableFilterWrapper
           filterComponent={
             <TableFilters
               amountText={isFetching ? 'Ladataan...' : `Löytyi ${count} kpl`}
```
tomimakela, line 453: This count seems to always say 0 in the map visualisation mode.

src/plotApplications/helpers.js
```diff
@@ -19,5 +23,59 @@ export const getContentApplicationListItem = (plotApplication: Object): Object =
...
+  };
+};
+
+/**
+ * Get application target features for geojson data
+ * @param {Object[]} targets
+ * @returns {Object[]}
+ */
+export const getApplicationTargetFeatures = (targets: Array<Object>): Array<LeafletFeature> => {
+  return targets.map((target) => {
```
tomimakela, line 41: What's passed in here is not targets but applications.

src/plotApplications/helpers.js
```diff
@@ -19,5 +23,59 @@ export const getContentApplicationListItem = (plotApplication: Object): Object =
...
+    return {
+      type: 'Feature',
+      geometry: {
+        coordinates: coordinates,
+        type: 'MultiPolygon',
+      },
+      properties: {
+        id: target.id,
+        feature_type: 'plotApplication',
+        identifier: target.targets[0].identifier,
```
tomimakela, line 62: This doesn't work. If the application is applying for more than one target, the map will visually show all of its areas but then link to listing based on the first target whenever any of them is clicked. If the topmost shape drawn has a different first target, the link will be entirely unrelated.

An array of features, one for each target (i.e. mapping "target.targets", actually "application.targets") should be returned here instead. (Whether the same target across different applications should get multiple overlaid features is another matter, I got 9 on top of each other which results in a rather opaque shape.)

tomimakela (changes requested):
Additional observations:

- The tests need to be amended a bit, as is evident from the failing automated checks.
- Everything added since the last review was put into a single commit. Sometimes that's still OK if everything is a part of the same feature, but in this case, for example the fixes alleviating the comments from the last review are now jumbled up with the rest of the changes. (They could be separated with history rewriting, but that's a bit cumbersome at this point).

vincit-matu (approved):
Nice stuff! 

The only comment I have is that the `any` type could probably be substituted with `unknown` where possible - as it's a bit stricter when using the typed variables and offers type checking, but is basically otherwise similar to `any` (which I try to avoid). But I see nothing _wrong_ per se anywhere, at least in this quick glance I had, as that's more of a preference/suggestion.

Edit: this also assumes the use of TypeScript 3

EmiliaMakelaVincit (approved):

PR #392, opened 2022-09-26
Enable editing of base applicant information check data

EmiliaMakelaVincit:
External service integration is still pending.

### DIFF
--- a/src/plotApplications/components/infoCheck/PlotApplicationApplicantInfoCheck.js
+++ b/src/plotApplications/components/infoCheck/PlotApplicationApplicantInfoCheck.js
@@ -1,0 +1,134 @@
+// @flow
+import React, {PureComponent} from 'react';
+
+import {Column, Row} from "react-foundation";
+import {getApplicantInfoCheckItems} from "../../helpers";
+import PlotApplicationApplicantInfoCheckModal from "./PlotApplicationApplicantInfoCheckModal";
+import PlotApplicationInfoCheckCollapse from "./PlotApplicationInfoCheckCollapse";
+import {connect} from "react-redux";
+import {
+  getInfoCheckAttributes,
+  getIsUpdatingInfoCheckData,
+  getWasLastInfoCheckUpdateSuccessfulData
+} from "../../selectors";
+import {getFieldOptions, getLabelOfOption} from "../../../util/helpers";
+import {editInfoCheckItem} from "../../actions";
+import {getUserFullName} from "../../../users/helpers";
+
+type Props = {
+  infoCheckAttributes: Attributes,
+  infoCheckData: Array<Object>,
+  section: Object,
+  identifier: string,
+  editInfoCheckItem: Function
+};
+
+type State = {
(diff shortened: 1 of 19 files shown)


src/plotApplications/components/infoCheck/PlotApplicationApplicantInfoCheck.js
```diff
@@ -0,0 +1,134 @@
...
+
+    const {
+      infoCheckAttributes,
+      infoCheckData,
+    } = this.props;
+
+    const infoCheckStatusOptions = getFieldOptions(infoCheckAttributes, 'state');
+
+    return (
+      <PlotApplicationInfoCheckCollapse headerTitle="Hakijan käsittelytiedot">
```
PasiVuohijoki, line 91: should this be translated also?

src/plotApplications/components/infoCheck/PlotApplicationInfoCheckForm.js
```diff
@@ -0,0 +1,151 @@
...
+      onClose
+    } = this.props;
+
+    return (
+      <form>
+        <Row>
+          <Column small={4}>
+            <FormField
+              setRefForField={this.setRefForFirstField}
+              fieldAttributes={get(attributes, 'preparer')}
```
PasiVuohijoki, line 78: Other places in mvj-ui these field paths and titles (here: 'preparer') are stored as enums/constants.

src/plotApplications/components/infoCheck/PlotApplicationTargetInfoCheck.js
```diff
@@ -0,0 +1,26 @@
...
+
+type State = {
+
+};
+
+class PlotApplicationTargetInfoCheck extends PureComponent<Props, State> {
+  render(): React$Node {
+    return (
+      <PlotApplicationInfoCheckCollapse className="collapse__secondary" headerTitle="Kohteen käsittelytiedot">
+
```
PasiVuohijoki, line 18: Does this need the closing tag since component does not have any children?

PasiVuohijoki (approved):
2 comments added. Nothing special/breaking though. Good job!

src/plotApplications/components/infoCheck/PlotApplicationInfoCheckForm.js
EmiliaMakelaVincit: Weeell... yes and no. I feel there isn't much consistency there; in the plot search draft we built on for instance, that hadn't been done before, and the field names are still inlined to this day. I can do so here if you think it's useful, though.

src/plotApplications/components/infoCheck/PlotApplicationTargetInfoCheck.js
EmiliaMakelaVincit: Not in its current state indeed, being largely just a placeholder for future development (I believe target info checks aren't even in the BE yet).

PR #393, opened 2022-09-28
Small fixes to lease statistic reports

juhakujala:
**1. Remove unecessary fixed title from LeaseStatisticReportForm.js**

When creating reports, the page always showed the "Vuokrausten tilastoraportti" as a title for the section, even though there
was something else selected as the report type.

Remove the title completely as we are showing the selected report title in the report output.

**2. LeaseStatisticReportForm: Render submit button based on report type**

Remove fixed value checks to determine which button (Luo raportti / Lähetä sähköpostiin) to show. Use the `is_async` from the API to render the correct form submit button for each report type.

### DIFF
--- a/src/leaseStatisticReport/components/LeaseStatisticReportForm.js
+++ b/src/leaseStatisticReport/components/LeaseStatisticReportForm.js
@@ -20,7 +20,6 @@ import {
 import {
   LeaseStatisticReportPaths,
   LeaseStatisticReportTitles,
-  LeaseInvoicingReportTypes,
 } from '$src/leaseStatisticReport/enums';
 import {
   getReportTypeOptions,
@@ -122,7 +121,6 @@ class LeaseStatisticReportForm extends PureComponent<Props, State> {
       reports,
       isFetchingReports,
       isFetchingReportData,
-      reportType,
       isSendingMail,
       options,
       isFetchingOptions,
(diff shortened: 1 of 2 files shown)


src/leaseStatisticReport/components/LeaseStatisticReportForm.js
```diff
@@ -168,15 +167,15 @@ class LeaseStatisticReportForm extends PureComponent<Props, State> {
                   </Column>
                 );
               })}
-              {((reportType !== LeaseInvoicingReportTypes.RENT_FORECAST && reportType !== LeaseInvoicingReportTypes.LEASE_STATISTIC) && fields && !isFetchingOptions) &&<Column small={3} style={{margin: '10px 0'}}>
+              {(!isAsync && fields && !isFetchingOptions) &&<Column small={3} style={{margin: '10px 0'}}>
```
PasiVuohijoki, line 170: Newline before the opening tag of Column would be nice but since there are quite many similar little things here and there, it would take too much time to fix all of those.

PasiVuohijoki (approved):
Nothing to say about the changes. The old codebase is quite unclean here and there, but let's not go into that.

src/leaseStatisticReport/components/LeaseStatisticReportForm.js
juhakujala: I agree that there would be a lot to fix in what comes to code formatting...

PR #395, opened 2022-10-07
Add special handling for applicant type

EmiliaMakelaVincit:
- Subsections within the top level applicant section will or won't be shown based on the selected applicant type
- The primary identifying detail (personal identity number or company ID) is attached into the metadata object of the section
- This identifying detail is used to label the collapse section when the application is viewed

Also includes Flow typing fixes (mostly permanent, but also some stop-gap temporary Object fallbacks) to all involved files.

### DIFF
--- a/src/plotApplications/components/PlotApplication.js
+++ b/src/plotApplications/components/PlotApplication.js
@@ -38,10 +38,13 @@ import {
 } from "../../plotSearch/selectors";
 import {getFieldAttributes} from "../../util/helpers";
 import ExternalLink from "../../components/links/ExternalLink";
-import {APPLICANT_SECTION_IDENTIFIER, TARGET_SECTION_IDENTIFIER} from "../constants";
+import {APPLICANT_MAIN_IDENTIFIERS, APPLICANT_SECTION_IDENTIFIER, TARGET_SECTION_IDENTIFIER} from "../constants";
 import {getApplicationRelatedPlotSearch, getIsFetchingApplicationRelatedPlotSearch} from "../selectors";
 import PlotApplicationTargetInfoCheck from "./infoCheck/PlotApplicationTargetInfoCheck";
 import PlotApplicationApplicantInfoCheck from "./infoCheck/PlotApplicationApplicantInfoCheck";
+import {ApplicantTypes} from "../enums";
+
+type OwnProps = {};
 
 type Props = {
   usersPermissions: UsersPermissionsType,
@@ -135,6 +138,10 @@ const SingleSectionItem = ({ section, answer, fieldTypes, plotSearch, topLevel, 
 };
 
 const SectionData = ({section, answer, topLevel = false, fieldTypes, plotSearch }) => {
+  if (!answer) {
+    return null;
+  }
+
   const title = section.title || "(tuntematon osio)";
 
   const Wrapper = topLevel ?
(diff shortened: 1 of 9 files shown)


PasiVuohijoki (approved):

src/plotApplications/helpers.js
```diff
@@ -309,45 +331,88 @@ export const prepareApplicationForSubmission = () => {
...
+              applicant.sections = _.pickBy(
+                _.cloneDeep(applicant.sections),
+                (subsection, sectionIdentifier: string) => [ApplicantTypes.BOTH, applicant.metadata.applicantType].includes(
+                  applicant.sectionRestrictions[sectionIdentifier]
+                )
+              );
+
+              try {
+                let identifier;
+                if (applicant.metadata.applicantType === ApplicantTypes.PERSON) {
```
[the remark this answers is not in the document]
EmiliaMakelaVincit: Not if written like that (since there's no overlap between the allowed values for applicantType and the section identifiers) but I think I get what you mean. Rather than defining simple string constants, I could do

```js
export const APPLICANT_MAIN_IDENTIFIER_SECTIONS = {
  [ApplicantTypes.COMPANY]: 'yrityksen-tiedot',
  [ApplicantTypes.PERSON]: 'henkilon-tiedot'
};

export const APPLICANT_MAIN_IDENTIFIER_FIELDS = {
  [ApplicantTypes.COMPANY]: 'y-tunnus',
  [ApplicantTypes.PERSON]: 'henkilotunnus'
}
```

and then do away with this if-else by just writing

```js
                  const sectionWithIdentifier = applicant.sections[APPLICANT_MAIN_IDENTIFIER_SECTIONS[applicant.metadata.applicantType]];
                  identifier = sectionWithIdentifier?.fields[APPLICANT_MAIN_IDENTIFIER_FIELDS[applicant.metadata.applicantType]]
                    || sectionWithIdentifier[0]?.fields[APPLICANT_MAIN_IDENTIFIER_FIELDS[applicant.metadata.applicantType]];
```
here instead. (With some additional intermediate variables for brevity probably. And the constant objects could also be combined even further.)

EmiliaMakelaVincit:
I restructured the constants as mentioned above, and also changed the collapse label logic for the applicants (because we didn't want to show personal identity numbers there). Could use a quick recheck.

src/plotApplications/components/PlotApplication.js
```diff
@@ -135,6 +138,10 @@ const SingleSectionItem = ({ section, answer, fieldTypes, plotSearch, topLevel,
 };
 
 const SectionData = ({section, answer, topLevel = false, fieldTypes, plotSearch }) => {
+  if (!answer) {
```
PasiVuohijoki, line 141: This 'answer' Variable is a bit confusing to me.

It is sometimes handled as an object (fe. SingleSectionItem -component there is `const fieldAnswer = answer.fields[field.identifier];`

Where as in row 187 it is passed as array :thinking:

PasiVuohijoki (commented):
I guess I don't have the real type of application template, so couldn't check this thoroughly. At least it does not crash anymore :+1:  :smile: 

The answer -variable was a bit confusing on PlotApplication.js as commented.

I try to find the correct application template and create a "real" answer/application from public ui and get back to this.

src/plotApplications/components/PlotApplication.js
EmiliaMakelaVincit: `answer` can both be an object (with `fields`, `sections` and possibly `metadata`) or an array of such objects. It's the latter only if the section it relates to can be duplicated automatically (i.e. top level target sections) or manually (i.e. top level applicant sections, or any other ones set in the form as such). Line 158 makes this distinction by reading `section.add_new_allowed`, and the handling inside the ternary branches should be consistent with that.

As long as the form itself hasn't been modified after the application was sent in, which should be the case given we now lock that when the application period has started, these should always agree with the form in the application data.

PasiVuohijoki:
I have now checked this with "latest" application and everything works fine :+1:

PR #396, opened 2022-10-07
Fix new plot search creation

EmiliaMakelaVincit:
- Fix crash due to an attempt to read a non-existing stage field
- Automatically set the "in preparation" stage when editing a newly created plot search

Also includes type fixes for the files involved.


src/plotSearch/components/plotSearchSections/basicInfo/BasicInfoEdit.js
```diff
@@ -52,21 +52,21 @@ import PlotSearchSiteEdit from './PlotSearchSiteEdit';
...
   formName: string,
-  isSaveClicked: Boolean,
+  isSaveClicked: boolean,
   usersPermissions: UsersPermissionsType,
+  decisionCandidates: Array<Object>,
+  hasUnidentifiedDecisions: boolean
 }
-
-// TODO
-// eslint-disable-next-line
```
PasiVuohijoki, line 69: Always good to get rid of these!

PR #397, opened 2022-10-10
Implement credit status integration for applicant info checks

EmiliaMakelaVincit:


src/creditDecision/components/CreditDecisionTemplate.js
```diff
@@ -6,13 +6,13 @@ import CreditDecisionHistory from '$src/creditDecision/components/CreditDecision
 import {ContactTypes} from '$src/contacts/enums';
 
 type Props = {
-  businessId?: String,
-  contactId?: String,
-  contactType: String,
-  nin?: String,
+  businessId?: string,
```
PasiVuohijoki, line 9: Is there anything else different between string and String but the looks? :thinking: 

I should propably then start to convert all the "capitalized" types also?

src/plotApplications/components/infoCheck/PlotApplicationApplicantInfoCheck.js
```diff
@@ -14,13 +14,21 @@ import {
...
+import {ApplicantTypes} from "../../enums";
+
+type OwnProps = {
+  section: Object,
+  identifier: string,
+  answer: Object
+};
 
 type Props = {
+  ...OwnProps,
```
PasiVuohijoki, line 27: Hmm. This is brilliant practice to separate the ui-related props from the object-related props coming from the BE.

I might need to refactor the CustomDetailedPlan to use same kind of approach. Much more readable component!

PasiVuohijoki (approved):

src/plotApplications/components/infoCheck/PlotApplicationApplicantInfoCheck.js
EmiliaMakelaVincit: Indeed. Well, I've mostly separated them just to get the right props exposed for the type coercion of the default export (I believe `connect`/HOC provided props would be marked as missing otherwise), but it does make it a bit more readable too.

PR #399, opened 2022-10-13
Fix links in clickable table rows

PasiVuohijoki:


EmiliaMakelaVincit (approved):
I'm not entirely certain that doing this for _all_ external links (even the ones outside of listings) is perfectly fine, but I can't think of a counterexample either, so I guess it shouldn't be a problem.

PR #402, opened 2022-11-15
Fix basis of rent calculator issues

juhakujala:
* Remove whitespace from `src/leases/enums.js`
* Fix issues in area unit and intended use
  * Add missing `LEASE2022` check
*  Move checkbox fields to their own row
  * Remove if clause from `locked_at` to always show `locked_at` field
* Fix copy to clipboard fields
  * Use it only for the `LEASE` and `LEASE2022` types
  * Update what fields are copied
* Add missing disabled attribute to fields with `locked_at`
* Fix rent titles
* Update `locked_at` text
* Sort basis of rent children by their id when fetching data

### DIFF
--- a/src/leases/components/leaseSections/rent/BasisOfRent.js
+++ b/src/leases/components/leaseSections/rent/BasisOfRent.js
@@ -427,10 +427,10 @@ const BasisOfRent = ({
               isFieldAllowedToRead(leaseAttributes, LeaseBasisOfRentsFieldPaths.AMOUNT_PER_AREA)
             }>
               {calculatorType === CalculatorTypes.TEMPORARY && <FormTextTitle enableUiDataEdit uiDataKey={getUiDataLeaseKey(LeaseBasisOfRentsFieldPaths.BASE_YEAR_RENT)}>
-                {'Vuokra/kuukausi'}
+                {'Vuokra/kk'}
               </FormTextTitle>}
               {calculatorType !== CalculatorTypes.TEMPORARY && <FormTextTitle enableUiDataEdit uiDataKey={getUiDataLeaseKey(LeaseBasisOfRentsFieldPaths.BASE_YEAR_RENT)}>
-                {LeaseBasisOfRentsFieldTitles.RENT}
+                {'Vuokra/vuosi'}
               </FormTextTitle>}
               {calculatorType === CalculatorTypes.TEMPORARY && <FormText>{!isEmptyValue(rent) ? `${formatNumber(rent)} €` : '-'}</FormText>}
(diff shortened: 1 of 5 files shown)


src/leases/components/leaseSections/rent/BasisOfRentEdit.js
```diff
@@ -1069,6 +1097,7 @@ class BasisOfRentEdit extends PureComponent<Props, State> {
                   fieldAttributes={getFieldAttributes(leaseAttributes, LeaseBasisOfRentsFieldPaths.TYPE)}
                   onChange={this.onChangeTypeOptions}
                   name={`${field}.type`}
+                  disabled={!!savedBasisOfRent && !!savedBasisOfRent.locked_at}
```
EmiliaMakelaVincit, line 1100: This is repeated quite a few times in here; could just be evaluated once somewhere higher up and then that variable used in each occurrence instead.

src/leases/components/leaseSections/rent/BasisOfRent.js
```diff
@@ -444,10 +427,10 @@ const BasisOfRent = ({
               isFieldAllowedToRead(leaseAttributes, LeaseBasisOfRentsFieldPaths.AMOUNT_PER_AREA)
             }>
               {calculatorType === CalculatorTypes.TEMPORARY && <FormTextTitle enableUiDataEdit uiDataKey={getUiDataLeaseKey(LeaseBasisOfRentsFieldPaths.BASE_YEAR_RENT)}>
-                {'Vuokra/kuukausi'}
+                {'Vuokra/kk'}
               </FormTextTitle>}
               {calculatorType !== CalculatorTypes.TEMPORARY && <FormTextTitle enableUiDataEdit uiDataKey={getUiDataLeaseKey(LeaseBasisOfRentsFieldPaths.BASE_YEAR_RENT)}>
-                {LeaseBasisOfRentsFieldTitles.RENT}
+                {'Vuokra/vuosi'}
```
EmiliaMakelaVincit, line 433: I kinda feel that these titles need a bit of standardisation. Either all of them in a component should be defined in the title enums (which has been historically the default for the project as far as I know), or all of them should be written here directly, but not a mix of those two. (Which was already the case prior to this PR I guess.)

src/leases/components/leaseSections/rent/BasisOfRentEdit.js
juhakujala: Yep, there's a lot of repetitive stuff in this file. Let's start from somewhere and I'll create a variable out of this 👍

src/leases/components/leaseSections/rent/BasisOfRent.js
juhakujala: There has been a mix of these two. I'll improve these a bit and convert these to ENUMS

PR #404, opened 2022-11-29
Improve plot application attachment handling

EmiliaMakelaVincit:
- The path parameter is now put into use now that it's exposed and set by the backend. The path defines the exact position the file is contained in in the answer data, which allows each occurrence of a certain field object to have unique file lists (i.e. two targets can have different reference attachments).
- Many quirks when the user aborts an edit of the application have been cleaned up. Files that are removed in the editor will now actually only be deleted when the edit is submitted, and newly uploaded files will also only be linked to the field during the save. Would-be linked files from applications that were never finished and submitted will also no longer show up the next time a new application editor (possibly for an unrelated set of targets) is opened.

### DIFF
--- a/src/plotApplications/helpers.js
+++ b/src/plotApplications/helpers.js
@@ -11,7 +11,6 @@ import {FormNames} from '../enums';
 import {
   getApplicationInfoCheckData,
   getCurrentEditorTargets,
-  getPendingUploads,
 } from './selectors';
 import {
   APPLICANT_MAIN_IDENTIFIERS,
@@ -116,7 +115,7 @@ export const reshapeSavedApplicationObject = (
 
   const result = getEmptySection();
 
-  const reshapeSingleSectionData = (section, answersNode) => {
+  const reshapeSingleSectionData = (section, answersNode, sectionPath) => {
     const data = getEmptySection();
 
     if (answersNode.metadata) {
@@ -124,7 +123,7 @@ export const reshapeSavedApplicationObject = (
     }
 
     section.subsections.forEach((subsection) => {
-      reshapeArrayOrSingleSection(subsection, data, answersNode);
+      reshapeArrayOrSingleSection(subsection, data, answersNode, `${sectionPath}.${subsection.identifier}`);
     });
 
     section.fields.forEach((field) => {
@@ -142,8 +141,10 @@ export const reshapeSavedApplicationObject = (
           }
           break;
         case 'uploadfiles':
+          // auto added attachment path goes only up to the section as field is already in its own field
+          value = attachments.filter((attachment) =>
+            attachment.field === field.identifier && attachment.path === sectionPath);
           extra_value = '';
-          value = attachments.filter((attachment) => attachment.field === field.identifier);
           break;
       }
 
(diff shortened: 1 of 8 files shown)


src/plotApplications/helpers.js
```diff
@@ -557,3 +562,8 @@ export const getSectionApplicantType = (state: RootState, section: FormSection,
 
   return formValueSelector(FormNames.PLOT_APPLICATION)(state, `${reduxFormPath}.metadata.applicantType`) || ApplicantTypes.UNSELECTED;
 };
+
+export const getFieldFileIds = (state: RootState, fieldPath: string): Array<number> => {
+  const fieldValue = formValueSelector(FormNames.PLOT_APPLICATION)(state, fieldPath);
+  return fieldValue?.value || [];
+}
```
PasiVuohijoki, line 569: missing semicolon

src/plotApplications/helpers.js
```diff
@@ -21,7 +21,7 @@ import {
 import type {LeafletFeature, LeafletGeoJson} from '$src/types';
 import {ApplicantInfoCheckTypes, ApplicantTypes, PlotApplicationInfoCheckExternalTypes} from './enums';
 import type {RootState} from '../root/types';
-import type {PlotApplicationFormValue, ApplicationFormSection, ApplicationFormState} from './types';
+import type {PlotApplicationFormValue, ApplicationFormSection, ApplicationFormState, UploadedFileMeta} from './types';
```
PasiVuohijoki, line 24: 'UploadedFileMeta' is defined but never used.

Same few rows earlier (14):

'getPendingUploads' is defined but never used.

PasiVuohijoki (commented):
Few type warnings and errors here and there. I started to point those out separately but I believe those are visible when making the flow checks so better check those yourself with that.

PasiVuohijoki (commented):
~I tried to upload reference attachments, but those never gotten seen on application. Not even when refreshed~

Nevermind, wrong branches used in backend. Works otherwise well, but when trying to download image (by clicking the link) results into not-authenticated response from api. But I guess that's an backend problem.

src/plotApplications/helpers.js
EmiliaMakelaVincit: Right, I forgot to clean up here after a bit of rearranging. Will fix.

EmiliaMakelaVincit:
> Few type warnings and errors here and there. I started to point those out separately but I believe those are visible when making the flow checks so better check those yourself with that.

If that's the case, I actually think that we're not seeing the same results. All files in the plotApplication directory show up with no type errors from Flow for me at the moment as far as I can see.

EmiliaMakelaVincit:
> Nevermind, wrong branches used in backend. Works otherwise well, but when trying to download image (by clicking the link) results into not-authenticated response from api. But I guess that's an backend problem.

It seems that the download URL only works when logged in into the admin site (which was the case for me, so I didn't notice anything odd). When the link is opened by regular browser navigation and not programmatically through JS, we're not sending the Authorization header anymore, so without both that and an existing session it looks just like a random unauthorized request. Hmm...

EmiliaMakelaVincit:
Attachment download should be fixed. I had already used the correct premade component in the edit view :D

PR #409, opened 2022-12-12
Remove hardcoded values from lease statistic reports

juhakujala:
* Remove all hardcoded checks for different types of fields based on their name
* Use value formatting provided by the backend
* Use the correct table sort function based on the is_numeric boolean

Depends on https://github.com/City-of-Helsinki/mvj/pull/463

### DIFF
--- a/src/leaseStatisticReport/components/LeaseInvoicingConfirmationReport.js
+++ b/src/leaseStatisticReport/components/LeaseInvoicingConfirmationReport.js
@@ -14,19 +14,22 @@ import FormText from '$components/form/FormText';
 import ExcelLink from '$components/excel/ExcelLink';
 import {
   getApiResponseResults,
-  formatDate,
-  formatNumber,
   hasPermissions,
   getLabelOfOption,
+  sortNumberByKeyAsc,
+  sortNumberByKeyDesc,
+  sortStringByKeyAsc,
+  sortStringByKeyDesc,
 } from '$util/helpers';
 import {
-  LeaseInvoicingReportPaths,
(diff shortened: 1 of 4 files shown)


src/leaseStatisticReport/components/LeaseInvoicingConfirmationReport.js
```diff
@@ -81,66 +84,31 @@ class LeaseInvoicingConfirmationReport extends PureComponent<Props, State> {
...
+          if (field.choices && value) {
+            outputValue = getDisplayName(field.choices, value);
+          } else if (field.format && value) {
+            outputValue = getFormattedValue(field.format, value);
+            isBold = field.format === LeaseStatisticReportFormatOptions.BOLD || field.format === LeaseStatisticReportFormatOptions.BOLD_MONEY;
+          }
+
+          return (
+            <FormText className='no-margin' style={{ whiteSpace: 'nowrap' }}>
+              {isBold ? <b>{outputValue}</b> : outputValue}
```
EmiliaMakelaVincit, line 105: `<strong>` is generally preferred for semantic reasons over `<b>` these days, I believe, but it works either way.

EmiliaMakelaVincit (approved):
Seems solid on code level. The new way the columns are defined is much less copy-pastey which is always a plus in my book 🙂

src/leaseStatisticReport/components/LeaseInvoicingConfirmationReport.js
juhakujala: I'll change this to `<strong>`

PR #410, opened 2022-12-15
Feature/custom detailed plan to plot search

PasiVuohijoki:
Custom detailed plan can now be selected as plot search target.
Backend branch: origin/feature-plotsearch-customdetailedplan

### DIFF
--- a/src/components/inputs/PlanUnitSelectInput.js
+++ b/src/components/inputs/PlanUnitSelectInput.js
@@ -4,7 +4,7 @@ import debounce from 'lodash/debounce';
 
 import AsyncSelect from '$components/form/AsyncSelect';
 import {getContentPlanUnitIdentifier} from '$src/plotSearch/helpers';
-import {fetchPlanUnitListWithIdentifiersList} from '$src/plotSearch/requestsAsync';
+import {fetchPlanUnitListWithIdentifiersList, fetchCustomDetailedPlanListWithIdentifiersList} from '$src/plotSearch/requestsAsync';
 
 type Props = {
   disabled?: boolean,
@@ -22,23 +22,31 @@ const PlanUnitSelectInput = ({
   onBlur,
   placeholder,
   value,
-}: Props) => {
-  const getPlanUnitOptions = (planUnitList: Array<Object>): Array<Object> =>
-    planUnitList
-      .map((plan_unit) => {
+}: Props): React$Node => {
+  const getPlanUnitOptions = (planUnitList: Array<Object>, customDetailedPlanList: Array<Object>): Array<Object> => {
+    return [...planUnitList, ...customDetailedPlanList]
+      .map((planUnit) => {
         return {
-          value: plan_unit.id,
-          label: getContentPlanUnitIdentifier(plan_unit),
+          value: planUnit.id,
+          label: getContentPlanUnitIdentifier(planUnit),
+          identifierType: planUnit.identifier_type,
         };
       });
+  };
+    
 
   const getPlanUnitList = debounce(async(inputValue: string, callback: Function) => {
     const planUnitList = await fetchPlanUnitListWithIdentifiersList({
-      search: inputValue,
       limit: 10,
+      search: inputValue,
     });
 
-    callback(getPlanUnitOptions(planUnitList));
+    const customDetailedPlanList = await fetchCustomDetailedPlanListWithIdentifiersList({
+      limit: 10,
+      search: inputValue,
+    });
+
+    callback(getPlanUnitOptions(planUnitList, customDetailedPlanList));
   }, 500);
   
   const input = {
@@ -52,7 +60,7 @@ const PlanUnitSelectInput = ({
     <AsyncSelect
       disabled={disabled}
       displayError={false}
-      getOptions={getPlanUnitList}
+      getOptions={(inputValue, callBack) => getPlanUnitList(inputValue, callBack)}
       input={input}
       isDirty={false}
       placeholder={placeholder}
--- a/src/plotSearch/components/plotSearchSections/basicInfo/PlotSearchSiteCustomDetailedPlan.js
+++ b/src/plotSearch/components/plotSearchSections/basicInfo/PlotSearchSiteCustomDetailedPlan.js
@@ -1,0 +1,293 @@
+// @flow
+import React, {PureComponent, Fragment} from 'react';
+import {connect} from 'react-redux';
+import {Row, Column} from 'react-foundation';
+
+import get from 'lodash/get';
+import {receiveCollapseStates} from '$src/plotSearch/actions';
+import {FormNames, ViewModes} from '$src/enums';
+import {Routes, getRouteById} from '../../../../root/routes';
+import Loader from '$components/loader/Loader';
+import LoaderWrapper from '$components/loader/LoaderWrapper';
+import Collapse from '$components/collapse/Collapse';
+import FormText from '$components/form/FormText';
+import FormTextTitle from '$components/form/FormTextTitle';
+import ExternalLink from '$components/links/ExternalLink';
+import {createPaikkatietovipunenUrl} from '$util/helpers';
+import {
+  formatDate,
+  getFieldOptions,
+  getLabelOfOption,
+} from '$util/helpers';
+import {
+  getAttributes,
+  getCollapseStateByKey,
+  getIsFetchingCustomDetailedPlanAttributes,
+} from '$src/plotSearch/selectors';
+import type {Attributes} from '$src/types';
+import {
+  PlotSearchFieldTitles,
+} from '$src/plotSearch/enums';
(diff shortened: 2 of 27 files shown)


src/plotSearch/components/plotSearchSections/basicInfo/PlotSearchSiteEditCustomDetailedPlan.js
```diff
@@ -0,0 +1,280 @@
...
+              {'Kaavayksikön laji'}
+            </FormTextTitle>
+            <FormText>{currentCustomDetailedPlan && getLabelOfOption(customDetailedPlanTypeOptions, currentCustomDetailedPlan.type.id) || '-'}</FormText>
+          </Column>
+          <Column small={6} medium={4} large={3}>
+            <FormTextTitle>
+              {'Kokonaisrakennusoikeus'}
+            </FormTextTitle>
+            <FormText>
+              {get(currentCustomDetailedPlan, 'rent_build_permission') || '-'}
```
EmiliaMakelaVincit, line 191: This is in square metres; the unit is missing.

src/plotSearch/components/plotSearchSections/basicInfo/PlotSearchSiteEditCustomDetailedPlan.js
```diff
@@ -0,0 +1,280 @@
...
+            </Row>
+            {usageDistributions.map((usageDistribution) => <Row key={usageDistribution.distribution}>
+              <Column small={12} medium={4} large={4}>
+                <FormText>
+                  {usageDistribution.distribution}
+                </FormText>
+              </Column>
+              <Column small={8} medium={6} large={6}>
+                <FormText>
+                  {usageDistribution.build_permission}
```
EmiliaMakelaVincit, line 264: This is also missing the unit.

src/plotSearch/components/plotSearchSections/basicInfo/PlotSearchSiteEditCustomDetailedPlan.js
```diff
@@ -0,0 +1,280 @@
...
+            </FormTextTitle>
+            <FormText>
+              {get(currentCustomDetailedPlan, 'preconstruction_estimated_construction_readiness_moment') || '-'}
+            </FormText>
+          </Column>
+          {infoLinks.length > 0 && (<Column small={12} medium={12} large={12} className="plot_search_target__info-links">
+            <Row>
+              <Column small={12} medium={4} large={4}>
+                <FormTextTitle>
+                  {PlotSearchFieldTitles.INFO_LINK_DESCRIPTION}
```
EmiliaMakelaVincit, line 206: 🔄 This component also suffers from having only some of the titles in the enum and some in the component source.

src/components/inputs/PlanUnitSelectInput.js
```diff
@@ -22,23 +22,40 @@ const PlanUnitSelectInput = ({
...
+          label: getContentPlanUnitIdentifier(customDetailedPlan),
+          identifierType: customDetailedPlan.identifier_type,
+        };
+      });
+
+    return [...planUnits, ...customDetailedPlans];
+  };
+    
+
   const getPlanUnitList = debounce(async(inputValue: string, callback: Function) => {
```
EmiliaMakelaVincit, line 49: I dunno if this was just for testing purposes, but the search string isn't passed through anymore and thus, rather unsurprisingly, actually searching for the identifier you're looking for isn't working.

EmiliaMakelaVincit (commented):
Functionally I didn't notice much amiss. Also as mentioned, several of the marked comments revolve around the older code that was copied over and is now under more scrutiny than before that we could now improve on.

src/plotSearch/components/plotSearchSections/basicInfo/PlotSearchSiteCustomDetailedPlan.js
```diff
@@ -0,0 +1,317 @@
...
+    );
+  }
+
+  componentWillUnmount() {
+    clearInterval(this.timer);
+  }
+
+  stopTimer = () => {
+    clearInterval(this.timer);
+  }
```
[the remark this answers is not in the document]
PasiVuohijoki: I think these are for the check if the plan unit has changed (aka. diverted from the master plan unit) But yeah, totally unnecessary for custom detailed plan!

src/plotSearch/helpers.js
```diff
@@ -130,9 +133,9 @@ export const getPlanUnitFromObjectKeys = (planUnit: Object, index: any): ?Object
 export const cleanTargets = (payload: Object, shouldFixTargetType: boolean): Object => {
   const plot_search_targets = payload.plot_search_targets.map(target => ({
     id: target.id,
-    plan_unit_id: target.plan_unit_id,
     target_type: shouldFixTargetType ? PlotSearchTargetType.SEARCHABLE : target.target_type,
-    info_links: target.info_links
+    info_links: target.info_links,
+    ...(target.plan_unit_id ? {plan_unit_id: target.plan_unit_id} : {custom_detailed_plan_id: target.custom_detailed_plan_id}),
```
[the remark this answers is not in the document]
PasiVuohijoki: Yeah, I know but backend goes boom if there is either id as null.

EmiliaMakelaVincit:
Something went wrong when pushing fixes in? When I try to see what changed, GitHub says: "There isn’t anything to compare. f813877fed35b7fb4d8eb65c8845d65f79902d3f and 5816f660ff1609058a978fa85296532f5dbde19d are identical."

src/plotSearch/components/plotSearchSections/basicInfo/PlotSearchSiteEdit.js
```diff
@@ -377,26 +416,67 @@ class PlotSearchSiteEdit extends Component<Props, State> {
...
 
+    if(toPlotSearch.identifierType === TargetIdentifierTypes.PLAN_UNIT && toPlotSearch) {
+      this.setState({
+        planUnitNew: toPlotSearch,
+        customDetailedPlanNew: null,
+      });
+ 
+      fetchPlanUnit(toPlotSearch);
+      this.changePlanUnitValue(toPlotSearch);
+      return;
```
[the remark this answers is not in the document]
EmiliaMakelaVincit: And with that, the early return is also not needed :p

src/plotSearch/reducer.js
```diff
@@ -207,17 +235,17 @@ const decisionCandidateReducer: Reducer<{ [planUnit: number]: Array<Object> }> =
...
     return newState;
   },
-  ['mvj/plotSearch/RESET_PLAN_UNIT_DECISIONS']: (state: Object) => {
-    return {};
-  }
-}, {});
+  ['mvj/plotSearch/RESET_PLAN_UNIT_DECISIONS']: () => {
+    return [];
+  },
+}, []);
```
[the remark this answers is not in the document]
EmiliaMakelaVincit: I see the shapes here were corrected, but two more points:
- The reducer type was `{ [planUnit: number]: Array<Object> }` before and now it's the less specific `Object`, that could be reverted too.
- The type hasn't been reverted in the expected tests base state.

src/components/inputs/PlanUnitSelectInput.js
EmiliaMakelaVincit: This is still broken. The new changes now filter the result list clientside, but that's unhelpful because only 10 plan units are fetched at a time – if you were looking for anything outside them, you won't get it. You have to add `searchInput` back into the parameters given for `fetchPlanUnitListWithIdentifiersList`.

src/plotSearch/components/plotSearchSections/basicInfo/PlotSearchSiteEditCustomDetailedPlan.js
EmiliaMakelaVincit: The unit was added but in a way that makes the default "no data" case never happen, as an empty field would still end up with a truthy string. It needs to be rearranged a bit, even if the flow in that case isn't as neat.

src/plotSearch/components/plotSearchSections/basicInfo/PlotSearchSiteEditCustomDetailedPlan.js
EmiliaMakelaVincit: This one is better if we can assume this is always set. I didn't notice before that the usage distribution fields lack the fallbacks, but if they should always be filled, then they're probably not needed.

src/plotSearch/helpers.js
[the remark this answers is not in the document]
EmiliaMakelaVincit: Fair enough, I guess. I do think at least with some request APIs/frameworks, fields with an undefined value would be skipped, but no idea if that'd apply here.

EmiliaMakelaVincit (approved):
The remaining bugs seem to now be resolved so this should be good enough for merging.

PR #411, opened 2023-01-09
Add target info check functionality

EmiliaMakelaVincit:
Implemented as part of the application view screen; a later PR will reorganise both the view and edit screens.

### DIFF
--- a/src/plotApplications/components/infoCheck/PlotApplicationTargetInfoCheck.js
+++ b/src/plotApplications/components/infoCheck/PlotApplicationTargetInfoCheck.js
@@ -1,18 +1,567 @@
 // @flow
 
 import React, {PureComponent} from 'react';
-import PlotApplicationInfoCheckCollapse from './PlotApplicationInfoCheckCollapse';
+import {Column, Row} from 'react-foundation';
+import flowRight from 'lodash/flowRight';
+import {connect} from 'react-redux';
+import get from 'lodash/get';
+import {FieldArray, getFormValues, reduxForm} from 'redux-form';
 
-type Props = {
+import FormField from '$components/form/FormField';
+import {formatDate, getFieldAttributes, isFieldAllowedToRead} from '$util/helpers';
+import Authorization from '$components/authorization/Authorization';
+import RemoveButton from '$components/form/RemoveButton';
+import FormTextTitle from '$components/form/FormTextTitle';
+import {ButtonColors} from '$components/enums';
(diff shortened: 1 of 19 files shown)


src/plotApplications/components/infoCheck/PlotApplicationTargetInfoCheck.js
```diff
@@ -1,18 +1,567 @@
 // @flow
 
-import React, {PureComponent} from 'react';
-import PlotApplicationInfoCheckCollapse from './PlotApplicationInfoCheckCollapse';
+import React, {Fragment, PureComponent} from 'react';
```
PasiVuohijoki, line 3: Fragment is defined but not used

src/plotApplications/requests.js
```diff
@@ -75,6 +75,41 @@ export const deleteUploadRequest = (
   );
 };
 
-export const editInfoCheckItemRequest = (payload: Object): Generator<any, any, any> => {
+export const editApplicantInfoCheckItemRequest = (payload: Object): Generator<any, any, any> => {
   return callApi(new Request(createUrl(`information_check/${payload.id}/`), {method: 'PATCH', body: JSON.stringify(payload)}));
 };
+
+export const editTargetInfoCheckItemRequest = (payload: Object): Generator<any, any, any> => {
+  return callApi(new Request(createUrl(`target_status/${payload.id}/`), {method: 'PATCH', body: JSON.stringify(payload)}));
```
PasiVuohijoki, line 83: I don't know if this a bug on my env or something, but currently the payload.id is undefined and it tries to call api with that resulting an error.

PasiVuohijoki:
Otherwise seems to work as should but pressing the save button will end up calling the BE without id.

src/plotApplications/requests.js
EmiliaMakelaVincit: Hmm... Can you check if the target statuses exist in the answer in the first place when fetched from the API? As it is right now, the form always shows up even if it's missing, as in the future there always should be a matching target status object for each target. If there isn't, then that application is in a broken state for one reason or another. Otherwise, there might be something I need to fix (though: works for me at the moment).

PasiVuohijoki: It was faulty form. the target was not dublicable.

PasiVuohijoki:
I tested this with faulty form. Works as it should now.

PR #413, opened 2023-01-31
Adds new rent type "index2022"

juhakujala:


EmiliaMakelaVincit (approved):
GitHub's diff algorithm didn't handle the indentation change in `ContractRentEdit.js` well, so reviewing any other changes that happened at the same time was a bit tricky, but I didn't spot anything amiss at least.

PR #415, opened 2023-02-13
Various application list view fixes

EmiliaMakelaVincit:


PasiVuohijoki:
Not related straight to this PR, but I needed to create bunch of new applications to make those go to the "second page" for testing purposes. Unfortunately submitting the application from "officials UI" resulted into this:
image

Not sure if my BE branch is correct. I'm using local test branch based on the develop branch where I have merged the "fix-geometry-field".

EmiliaMakelaVincit:
That's strange, it works for me with the same backend version. The trace itself seems a bit weird as it seems as if it's trying to find the "vahvistukset" section, a root level section, from under some other section instead (because there's two hits on helpers.js:195).

I think that's after the application has been successfully saved and it's trying to go to the regular view mode; is that page still broken if refreshed? Maybe something weird happened to the form itself during submission which causes the form and the application to not match anymore.

src/components/table/TableFilters.js
```diff
@@ -6,7 +6,7 @@ import CheckboxInput from '$components/inputs/CheckboxInput';
 
 type Props = {
   alignFiltersRight?: boolean,
-  amountText: string,
+  amountText: React$Node,
```
PasiVuohijoki, line 9: Couldn't this be string?

src/plotApplications/components/PlotApplicationsListPage.js
```diff
@@ -441,20 +429,15 @@ class PlotApplicationsListPage extends PureComponent<Props, State> {
...
+    const count = getApiResponseCount(isMap ? plotApplicationsMapData : plotApplicationsListData);
     const columns = this.getColumns();
 
     let amountText;
     if (isMap && isFetchingByBBox || isTable && isFetching) {
       amountText = 'Ladataan...';
     } else {
       if (isMap && Number(getUrlParams(search)?.zoom || 0) < MAX_ZOOM_LEVEL_TO_FETCH_LEASES) {
-        amountText = '';
+        amountText = <>&nbsp;</>;
```
PasiVuohijoki, line 440: Oh, this is why the prop type is ReactNode. Did the ' ' -string break something?

PasiVuohijoki (approved):
Everything seems to be good on the code. I left this one "nitpick/i wonder why" -comment, but I'm quite sure that there is explanation for that. :+1:

src/plotApplications/components/PlotApplicationsListPage.js
EmiliaMakelaVincit: A whitespace-only string gets condensed to nothing in HTML (without a non-default `white-space` CSS rule anyway) which meant the text element became 0 pixels tall in this case, thus leading to the inconsistent map positioning in the map view depending on the zoom level.

PasiVuohijoki: Ahh, the same thing that with line-breaks. I got it. :+1:

PR #417, opened 2023-02-16
Clean up code in plot searches and applications

EmiliaMakelaVincit:
- Fix all found Flow errors
- Fix all found eslint errors
- Convert remaining relative imports to absolute ones
- Replace remaining React$AbstractComponents with React$ComponentTypes

----

I thought it'd be better to separate these into their own PR so that if you happen to return to work on this UI before I have any actually complete other features to throw in, these fixes can already be used as a foundation for both.

### DIFF
--- a/src/components/search/SearchContainer.js
+++ b/src/components/search/SearchContainer.js
@@ -6,7 +6,10 @@ type Props = {
   onSubmit: Function,
 }
 
-const SearchContainer = ({children, onSubmit}: Props) => 
+const SearchContainer = ({children, onSubmit}: Props): React$Node =>
+  // TODO: determine which components actually have an use for onSubmit and consider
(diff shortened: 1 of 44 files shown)


PasiVuohijoki (approved):
No punakynä at all. Good job cleaning everything! Is the React$AbstractComponent incorrect in all the cases we have used it and we should change it to React$ComponentType in everywhere? Could that be done like find & replace style? Although I guess there is not much of those left anymore now that you have gone through the plotApplication folder.

EmiliaMakelaVincit:
I'm still not sure if it's _the_ correct type to use, but I believe it is more accurate to what we actually need there indeed. Easier to work with too since it doesn't have that second type parameter that we just always handwaved away with AbstractComponent.

PR #419, opened 2023-03-21
Change build permission unit labels from sqm to floor-sqm

EmiliaMakelaVincit:
Taiga issue \#1729. Quite a trivial issue, the only thing to check really is whether I actually caught all the places where this needed to be changed.


PasiVuohijoki (approved):
Couldn't find any missed cases. :+1:

PR #420, opened 2023-04-03
Add plot application export modal

EmiliaMakelaVincit:
The exact location of the export button might still have to be discussed, but all the functionality is there now.

### DIFF
--- a/src/plotApplications/components/exportModal/PlotSearchExportModal.js
+++ b/src/plotApplications/components/exportModal/PlotSearchExportModal.js
@@ -1,0 +1,261 @@
+// @flow
+import React, {Component} from 'react';
+import {flowRight} from 'lodash/util';
+import {change, formValueSelector, reduxForm} from 'redux-form';
+import {connect} from 'react-redux';
+
+import Modal from '$components/modal/Modal';
+import ModalButtonWrapper from '$components/modal/ModalButtonWrapper';
+import Button from '$components/button/Button';
+import {ButtonColors} from '$components/enums';
+import FormField from '$components/form/FormField';
+import {FieldTypes} from '$src/enums';
+import FileDownloadButton from '$components/file/FileDownloadButton';
+import createUrl from '$src/api/createUrl';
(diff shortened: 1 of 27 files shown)


src/plotApplications/components/exportModal/PlotSearchExportModal.js
```diff
@@ -0,0 +1,261 @@
...
+        case ExportModes.APPLICATIONS_BY_APPLICANT_PDF:
+          guideText = <p>Tulostus sisältää .pdf-muodossa yhden hakijan kaikki hakemukset liitteineen.</p>;
+          columns = [
+            ModalDetailColumns.CHECKBOX,
+            ModalDetailColumns.APPLICANT,
+            ModalDetailColumns.IDENTIFIER,
+            ModalDetailColumns.TARGET,
+          ];
+          if (selectedItemIds.length > 0) {
+            downloadUrl = createUrl(`target_status_pdf/?targets=${selectedItemIds.join(',')}`);
```
PasiVuohijoki, line 169: is this filter correct? should this be filtered by applicant in url?

src/plotApplications/components/exportModal/PlotSearchExportModal.js
PasiVuohijoki, line 1: This was a bit confusing at first sight.

The filtering is basically done UI-side and then we use the targets -parameter with the pre-filtered list of target id:s to fetch the PDF:s from the backend right?

I was just wondering, that could the filtering be done backend side as there is FilterView -mixin in use. So we could put the parameters on url. Like `target_status_pdf/?applicant=<selected_applicant_id_here>`

I'm pretty sure there is a good reason for this, but just curious to know.

PasiVuohijoki (approved):
Checked the functionality in local environment and everything worked as should.
Multiple flow-issues fixed, good!
I'm sure that the thing I mentioned in the comments about filtering the targets to export in fe-side instead of using the neat BE-filters is something that is discussed in dailies, but couldn't remember anymore.

PR #421, opened 2023-04-19
Add the reservation identifiers modal

EmiliaMakelaVincit:


src/components/form/FieldTypeSelect.js
```diff
@@ -106,7 +108,8 @@ class FieldTypeSelect extends PureComponent<Props, State> {
           isDisabled={disabled}
           filterOption={filterOption}
           id={name}
-          onBlue={this.handleBlur}
+          isLoading={isLoading}
+          onBlur={this.handleBlur}
```
PasiVuohijoki, line 112: wow :D

src/plotSearch/components/reservationIdentifiers/ReservationIdentifiersModal.js
```diff
@@ -0,0 +1,381 @@
...
+
+    if (!reservationIdentifierUnitLists) {
+      return '...';
+    }
+
+    const type = reservationIdentifierUnitLists.types.find((candidate) => candidate.id === options.type);
+    const municipality = reservationIdentifierUnitLists.municipalities.find((candidate) => candidate.id === options.municipality);
+    const district = reservationIdentifierUnitLists.districts.find((candidate) => candidate.id === options.district);
+
+    return `${type.identifier || '??'}${municipality.identifier || '?'}${padStart(district.identifier || '??', 2, '0')}`;
```
PasiVuohijoki, line 210: Type can be undefined if backend is failing.

PasiVuohijoki (commented):
IDE, Linter, Flow and I are happy with typing fixes.

As mentioned in the comments, I managed to crash the UI when testing with wrong BE. Should there be some integrity check of data so that it wont crash UI if BE is failing?

src/plotSearch/components/reservationIdentifiers/ReservationIdentifiersModal.js
EmiliaMakelaVincit: Hmm, I thought that shouldn't be possible. The saga that fetches the three lists shouldn't trigger the action to set them into the reducer if any of them couldn't be fetched, but it's possible I made a mistake there somewhere.

EmiliaMakelaVincit: Right, it wasn't because of that but because the plot search targets were missing the newly added fields. While it probably won't happen with the correct BE, I'll fix the fallbacks to also trigger if any unit was not found in the lists, because that's what they were supposed to do from the outset.

PasiVuohijoki (approved):

PasiVuohijoki:
Good to go!

PR #423, opened 2023-05-08
Implemented hidden and fractional field type.

PasiVuohijoki:
Added support for form field default value.

### DIFF
--- a/src/components/form/FieldTypeFractional.js
+++ b/src/components/form/FieldTypeFractional.js
@@ -1,0 +1,71 @@
+// @flow
+import React from 'react';
+import classNames from 'classnames';
+
+type Props = {
+  disabled: boolean,
+  displayError: boolean,
+  input: Object,
+  isDirty: boolean,
+  placeholder?: string,
+  setRefForField?: Function,
+}
+
+const NUMERATOR = 0;
+const DENOMINATOR = 1;
+const SPLITTER = ' / ';
+const FieldTypeFractional = ({
+  disabled = false,
+  displayError = false,
+  input: {name, onChange, value},
+  isDirty = false,
+  setRefForField,
+}: Props): React$Node => {
+
+  const handleSetReference = (element: any) => {
+    if(setRefForField) {
+      setRefForField(element);
+    }
+  };
+
+  const changeHandler = (newValue: string, fieldType: number): void => {
+    const values = value.split(SPLITTER);
+    values[fieldType] = newValue;
+    onChange(`${values[NUMERATOR] || ''}${SPLITTER}${values[DENOMINATOR] || ''}`);
+  };
+
+  const parseValue = (fieldType: number): string => {
+    if (!value) {
+      return '';
+    }
+    const values = value.split(SPLITTER);
+    return values[fieldType];
+  };
+
+  return (
+    <div className={classNames('form-field__fractional', {'has-error': displayError}, {'is-dirty': isDirty})}>
+      {/* numerator / fin: osoittaja */}
+      <input
+        ref={handleSetReference}
+        className="form-field__fractional-input"
+        id={`${name}-numerator`}
+        value={parseValue(NUMERATOR)}
+        onChange={(e) => changeHandler(e.target.value, NUMERATOR)}
+        disabled={disabled}
+        type="number"
+      />
+      <span className="form-field__fractional-divider"> / </span>
+      {/* denominator / fin: nimittäjä */}
+      <input
+        className="ApplicationFractionField__field"
+        id={`${name}-denominator`}
+        value={parseValue(DENOMINATOR)}
+        onChange={(e) => changeHandler(e.target.value, DENOMINATOR)}
+        disabled={disabled}
+        type="number"
+      />
+    </div>
+  );
+};
+
+export default FieldTypeFractional;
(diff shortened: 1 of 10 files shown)


src/components/form/FieldTypeFractional.js
```diff
@@ -0,0 +1,71 @@
...
+    if(setRefForField) {
+      setRefForField(element);
+    }
+  };
+
+  const changeHandler = (value: string, fieldType: number): void => {
+    const values = value.split(SPLITTER);
+    values[fieldType] = value;
+    onChange(`${values[NUMERATOR] || ''}${SPLITTER}${values[DENOMINATOR] || ''}`);
+  };
```
EmiliaMakelaVincit, line 35: Something here is wonky as this component didn't react to my inputs correctly. When I typed into the numerator field, the denominator field cleared itself, and when I typed into the denominator field, the same value appeared into both of them.

src/components/form/FieldTypeFractional.js
```diff
@@ -0,0 +1,71 @@
...
+    <div className={classNames('form-field__fractional', {'has-error': displayError}, {'is-dirty': isDirty})}>
+      {/* numerator / fin: osoittaja */}
+      <input
+        ref={handleSetReference}
+        className="form-field__fractional-input"
+        id={`${name}-numerator`}
+        value={parseValue(NUMERATOR)}
+        onChange={(e) => changeHandler(e.target.value, NUMERATOR)}
+        disabled={disabled}
+        type="text"
```
EmiliaMakelaVincit, line 55: Should be of type number, same with the other field.

src/components/form/FormField.js
```diff
@@ -71,6 +75,7 @@ const Types = {
   [FieldTypeOptions.REFERENCE_NUMBER]: 'text',
   [FieldTypeOptions.STRING]: 'text',
   [FieldTypeOptions.TEXTAREA]: 'text',
+  [FieldTypeOptions.FRACTIONAL]: 'text',
```
EmiliaMakelaVincit, line 78: I'm not sure if this does anything, though I don't fully understand the function of this enum in the first place. It seems to be only used to find out which value to pass down to the specific form field component as the `type` prop, but `FieldTypeFractional` doesn't make use of that prop. The fields that were previously defined here all already have "text" as the default value for that prop or don't use it at all either, so it's more or less of a noop?

src/plotSearch/components/plotSearchSections/application/ApplicationPreviewSection.js
```diff
@@ -171,6 +184,30 @@ class ApplicationPreviewSection extends PureComponent<Props> {
...
+      case 'hidden':
+        fieldSpecificComponents = <FormField
+          name={fakeFieldId}
+          fieldAttributes={{
+            type: 'hidden',
+            required: false,
+            read_only: true,
+            value: field.default_value,
+          }}
+          readOnlyValueRenderer={() => {
```
EmiliaMakelaVincit, line 196: I'd probably actually want to show the field label here like with other fields or otherwise indicate what the text is for, just for clarity.

src/plotSearch/components/plotSearchSections/application/ApplicationPreviewSection.js
EmiliaMakelaVincit, line 1: It might make more sense to drop them while generating the template forms, but the sections that never get used (like Yrityksen tiedot for a natural person only plot search) should probably not be shown here at all. Though it could be useful to show the conditionality for the other sections even in the usual case, which is something I didn't think of at all when first implementing that functionality on this side. Thoughts?

src/components/form/FieldTypeFractional.js
PasiVuohijoki: Variable name collision.. value is a prop for this component and also a parameter in this changeHandler function... fixed..

src/components/form/FormField.js
PasiVuohijoki: yah. can be removed.

src/plotSearch/components/plotSearchSections/application/ApplicationPreviewSection.js
PasiVuohijoki: Am I lazy if I suggest removing the "company" fields whenever we are generating the "person" -forms and vice versa in backend? :thinking:

src/plotSearch/components/plotSearchSections/application/ApplicationPreviewSection.js
PasiVuohijoki: I see. Good point!

src/plotApplications/components/PlotApplicationSubsection.js
```diff
@@ -276,6 +277,14 @@ const ApplicationFormSubsectionFields = connect(
             })),
           };
           break;
+        case 'hidden':
+          if(field.identifier === 'hakija') {
```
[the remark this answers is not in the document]
PasiVuohijoki: done

src/components/form/FieldTypeFractional.js
PasiVuohijoki: done

src/plotSearch/components/plotSearchSections/application/ApplicationPreviewSection.js
EmiliaMakelaVincit: Yeah, as discussed in the daily, let's remove them from the templates instead.

EmiliaMakelaVincit (approved):

PR #424, opened 2023-05-11
Implement area search list

EmiliaMakelaVincit:


PasiVuohijoki:
Not a bug, but something to think about.

Now I have a situation where there is area searches created, but no applications for those. So ie. I have closed the application page before sending the application.

Now the page lists those also. Maybe would be wise to filter those out. We have discussed to have some sort of garbage cleaner script which deletes area searches without applications once in a week or something like that.

PasiVuohijoki (approved):
Listing and filters works like should. I'd say we can merge this and think if we should filter the "non-application" area searches and implement it when merging the map portion of this..

EmiliaMakelaVincit:
Yeah, the incomplete searches do indeed show up on the list for now. Hiding them is essentially just more filtering, so it needs backend support to be implemented properly; I should pester Niko to make it so the backend either pre-emptively always excludes them, or alternatively has some kind of filter parameter that the UI can provide to remove them manually.

EmiliaMakelaVincit:
A new BE commit that hides them from view has now been added to address that issue.

PR #431, opened 2023-05-30
Area search detail page

EmiliaMakelaVincit:
Selected area minimap dimensions are not included as I couldn't figure out how to make them work. Feel free to point out the mistake in my approach if you're able to see the problem :)


PasiVuohijoki (approved):
Tried to tinker with the maps, but nothing mentionable results gained. :+1: 

Otherwise seems solid.

PR #432, opened 2023-06-21
Add area search information check editor

EmiliaMakelaVincit:
* The first commit contains the bulk of the refactoring of general application components (that apply to both plot search and area search applications). I suggest reviewing the commits individually for easier reading if possible, or I can split the PR into two as well (one for refactoring only, one for the rest) if that'd be helpful.
* As of right now, the backend isn't completely ready yet: the patch doesn't go through if no comment is entered.


PasiVuohijoki (approved):
Nothing to complain. :+1:

PR #433, opened 2023-06-22
Implement the preparer and lessor change modal in area search list

EmiliaMakelaVincit:


PasiVuohijoki (approved):
Does what it promises :+1:

PR #437, opened 2023-08-29
Plot search issues

EmiliaMakelaVincit:


PasiVuohijoki (approved):
Looks good. The solution to hide hidden fields made me laugh. I have done a bit overthinking there? :laughing:

EmiliaMakelaVincit:
I mean, it seemed fine to me when it was originally implemented. Back then it was just misunderstood that they'd provide more value when they are visible.

PR #440, opened 2023-09-08
Fix address search input

NC-jsAhonen:
Previously, the address search field could not find address data so that it would show the street number. Moreover, it used several endpoints to search first for addresses, and if it could not find them, it used streets as a fallback. This fix makes the search results show the street number and the municipality, and simplifies the code by using only one endpoint. For this reason, a lot of logic is removed.

### DIFF
--- a/src/components/address-search/AddressSearchInput.js
+++ b/src/components/address-search/AddressSearchInput.js
@@ -233,7 +233,7 @@ class AddressSearchInput extends Component<Props, State> {
   searchByKeyword = debounce((input: string) => {
     const fetchByKeyword = (language: Language) => {
       const url = `${SERVICE_MAP_URL}/search/?${stringifyQuery({
-        page_size: 4,
+        page_size: 25,
         type: 'address',
         input: input,
         language: language,
--- a/src/components/address-search/AddressSearchInput.js
+++ b/src/components/address-search/AddressSearchInput.js
@@ -15,16 +15,6 @@ const SERVICE_MAP_URL = 'https://api.hel.fi/servicemap/v2';
 type Language = 'fi' | 'sv';
 const MINIMUM_SEARCH_STRING = 4
 
-type Street = {
-  id: number,
-  language: Language,
-  municipality: string,
-  name: {
-    fi: string,
-    sv: string,
-  },
-}
-
 type Address = {
   object_type: string,
   name: {
@@ -71,7 +61,6 @@ type State = {
   isLoading: boolean,
   menuOpen: boolean,
   selectedAddress: ?Address,
-  streets: Array<Street>,
   value: ?string,
 }
 
@@ -87,7 +76,6 @@ class AddressSearchInput extends Component<Props, State> {
     isLoading: false,
     menuOpen: false,
     selectedAddress: null,
-    streets: [],
     value: '',
   }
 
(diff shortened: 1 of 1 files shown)


NC-jsAhonen:
The search results are quirky at times, and there are two causes for them:

- The endpoint does not find matches perfectly. Sometimes, for example, it might ignore double consonants (for example, if you write "takka", the response might include "taka" instead)
- Sometimes, the search result list is populated with results that are not from the latest query. I don't know, how to make it ignore the earlier queries and just use the latest one.

src/components/address-search/AddressSearchInput.js
```diff
@@ -9,34 +9,43 @@ import Loader from '$components/loader/Loader';
 import LoaderWrapper from '$components/loader/LoaderWrapper';
 import {stringifyQuery} from '$src/api/createUrl';
 import {KeyCodes} from '$src/enums';
-import {findFromOcdString, hasNumber} from '$util/helpers';
+import {findFromOcdString} from '$util/helpers';
 
 const SERVICE_MAP_URL = 'https://api.hel.fi/servicemap/v2';
 type Language = 'fi' | 'sv';
+const MINIMUM_SEARCH_STRING = 4
```
antti-honkanen, line 16: Is 4 really the best? For example GitLab uses 3 as min value for searching.

src/components/address-search/AddressSearchInput.js
```diff
@@ -308,14 +245,14 @@ class AddressSearchInput extends Component<Props, State> {
...
 
       const fiResults = await fiResponse.json();
-      const svResults = await svResponse.json();
+      // const svResults = await svResponse.json();
 
       return [
-        ...fiResults.results.map((street) => ({...street, language: 'fi'})),
-        ...svResults.results.map((street) => ({...street, language: 'sv'})),
+        ...fiResults.results.map((address) => ({...address, language: 'fi'})),
+        // ...svResults.results.map((address) => ({...address, language: 'sv'})),
```
antti-honkanen, line 255: Why having code for Swedish results, but not use them? Should consider removing this from the codebase.

src/components/address-search/AddressSearchInput.js
```diff
@@ -338,175 +275,35 @@ class AddressSearchInput extends Component<Props, State> {
...
-
-        newState.isLoading = false;
-        this.setState(newState);
-      })
-      .catch((error) => {
-        this.setState({isLoading: false});
-        console.error(`Failed to fetch streets with error ${error}`);
-      });
-  }, 300);
+  }, 750)
```
antti-honkanen, line 278: Magic number. Should use variable for this to give some context.

Is 750 milliseconds a bit too long debounce time? Would 500 ms (half a second) be better option? I know it is a tough balance between wait time and performing searches, but still this needs to be carefully decided. Waiting too long causes delays in filling to form, because user has to wait also the results and not only the debounce time.

src/components/address-search/AddressSearchInput.js
NC-jsAhonen: It probably doesn't matter very. I changed it to 3. The main point is to not make queries with extremely short inputs.

src/components/address-search/AddressSearchInput.js
NC-jsAhonen: Removed

src/components/address-search/AddressSearchInput.js
NC-jsAhonen: Yes, I experimented with this. I had to increase it from 300 anyway, because the search behaved a bit strange. But it seems to be fixed now. I set ti to 500ms and assigned it to variable.

PR #443, opened 2023-10-09
Implement a full-fledged plot search form section editor

EmiliaMakelaVincit:
The editor allows the user to specify all data necessary for setting up a customisable, but structurally coherent form for any specific plot search. This includes automatic identifier specification as well as protection of specific operationally important fields and sections containing said fields.


PasiVuohijoki (commented):
Should there be somekind of option to create totally new application form?

image

There could be option "Luo uusi lomake" and there would be only the needed fields for UI to work.

PasiVuohijoki (approved):
Codewise and functionwise nothing to complain. I guess this can be merged and let's focus on other areas and decide the "add new form" -option later?

EmiliaMakelaVincit:
Yeah, I think that hasn't been in the scope for now. Copying an existing form from another search was discussed at some point, though, but that was also left for some other day.

PR #444, opened 2023-10-13
Plot search form validations

EmiliaMakelaVincit:
Adds validators for select application form fields, and also fixes some area search specs page validations that were fixed on the public side earlier.

### DIFF
--- a/src/application/formValidation.js
+++ b/src/application/formValidation.js
@@ -1,0 +1,176 @@
+// @flow
+import {parseISO} from 'date-fns';
+import {get, set} from 'lodash/object';
+
+import {CONTROL_SHARE_FIELD_IDENTIFIER} from '$src/application/constants';
+
+const PERSONAL_IDENTIFIER_CHECK_CHAR_LIST = '0123456789ABCDEFHJKLMNPRSTUVWXY';
+// from the rightmost digit to the leftmost
+const COMPANY_IDENTIFIER_CHECKSUM_MULTIPLIERS = [2, 4, 8, 5, 10, 9, 7];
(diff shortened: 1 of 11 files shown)


src/application/formValidation.js
```diff
@@ -0,0 +1,176 @@
...
+export const personalIdentifierValidator = (value: any, error?: string): ?string => {
+  if (value === '') {
+    return;
+  }
+
+  if (typeof value !== 'string') {
+    return error || 'Virheellinen henkilötunnus';
+  }
+
+  const result = /^(\d{6})([-+ABCDEFUVWXY])(\d{3})([0-9ABCDEFHJKLMNPRSTUVWXY])$/.exec(value.toUpperCase());
```
PasiVuohijoki, line 20: I just realised, that backend does not have this touppercase -operation and therefore this could pass the ui validation but fail at the be-side.

PasiVuohijoki (approved):
I wrote a ticket to fix the be-side ssn-validation to transfer the entry to uppercase to have similar results than fe-side validation.

src/application/formValidation.js
EmiliaMakelaVincit:
```py
            FieldRegexValidator(
                "^[0-9]{6}[+Aa-][0-9]{3}[A-z0-9]$", "invalid_ssn", "henkilotunnus"
            ),
```

To me it does seem it accepts uppercase characters... though incorrectly (A-z includes these characters besides the intended letters: [\\]^_\`). It also doesn't seem to accept the recently added additional separator characters.

PR #445, opened 2023-11-01
807 history tree with plotsearch

NC-jsAhonen:
Historiapuuhun sisällytetään tonttihaut, tonttihakemukset ja aluehakemukset. Ne näytetään, ja niitä voi manuaalisesti lisäillä ja poistaa. Automaattisesti lisätyt tiedot näytetään myös, mutta niitä ei voi poistaa.

Toimii yhdessä mvj backendin samannimisen `807-history-tree-with-plotsearch` branchin kanssa.


src/components/inputs/LeaseSelectInput.js
```diff
@@ -22,27 +31,86 @@ const LeaseSelectInput = ({
...
 
-  const getLeases = debounce(async(inputValue: string, callback: Function) => {
+    plotSearches = plotSearches
+      .filter(filterOutExisting)
+      .map((plotSearch) => {
+        const maxLength = 24
+        const displayName = plotSearch.name.length > maxLength ? plotSearch.name.substr(0,maxLength) + "..." : plotSearch.name
+        return {
+          value: plotSearch.id,
+          label: `${displayName}, Haku`,
```
henrinie-nc, line 59: Get "Haku" from translations

src/components/inputs/LeaseSelectInput.js
```diff
@@ -22,27 +31,86 @@ const LeaseSelectInput = ({
...
+          content_type_model: "plotsearch"
+        };
+      });
+
+    plotApplications = plotApplications
+      .filter(filterOutExisting)
+      .map((plotApplication) => {
+        return {
+          value: plotApplication.id,
+          label: `${plotApplication.application_identifier}, Hakemus`,
```
henrinie-nc, line 70: Get Hakemus from translations, or translate it

src/components/inputs/LeaseSelectInput.js
```diff
@@ -22,27 +31,86 @@ const LeaseSelectInput = ({
...
+          content_type_model: "targetstatus"
+        };
+      });
+
+    areaSearches = areaSearches
+      .filter(filterOutExisting)
+      .map((areaSearch) => {
+        return {
+          value: areaSearch.id,
+          label: `${areaSearch.identifier}, Aluehakemus`,
```
henrinie-nc, line 81: This should also be translatable

src/components/inputs/LeaseSelectInput.js
```diff
@@ -22,27 +31,86 @@ const LeaseSelectInput = ({
...
+          type: 'related_plot_application',
+          content_type_model: "areasearch"
+        };
+      });
+
+      return [...leases, ...plotSearches, ...plotApplications, ...areaSearches].sort((a, b) => (a.label && b.label) && a.label > b.label ? 1 : -1).slice(0,10)
```
henrinie-nc, line 87: How does this sorting work, does it sort by labels of the history items? And then what is the slice for?

src/components/inputs/LeaseSelectInput.js
```diff
@@ -22,27 +31,86 @@ const LeaseSelectInput = ({
...
+
+  const getHistoryItems = debounce(async(inputValue: string, callback: Function) => {
     const leases = await fetchLeases({
       succinct: true,
       identifier: inputValue,
       limit: 10,
     });
+    
+
+    const plotSearches = await fetchPlotSearches({
```
henrinie-nc, line 98: How about using Promise.all() here?

src/leases/components/leaseSections/summary/LeaseHistory.js
```diff
@@ -0,0 +1,197 @@
...
+
+type Props = {
+  currentLease: Lease,
+  leaseAttributes: Attributes,
+}
+
+type State = {
+  currentLease: Lease,
+  leaseAttributes: Attributes,
+  relatedLeasesFrom: Array<Object>,
```
henrinie-nc, line 28: Is it possible to use something like Array<Lease> here? Not familiar with this typing system

src/leases/components/leaseSections/summary/LeaseHistory.js
```diff
@@ -0,0 +1,197 @@
...
+    leaseAttributes: null,
+    relatedLeasesFrom: [],
+    relatedLeasesTo: [],
+    stateOptions: [],
+  }
+
+  static getDerivedStateFromProps(props: Props, state: State) {
+    const newState = {};
+
+    if(props.leaseAttributes !== state.leaseAttributes) {
```
henrinie-nc, line 45: I recommend formatting the code as in this PR I see multiple different conventions like `if()` and `if ()`, also some lists have space between items and some not

src/leases/components/leaseSections/summary/LeaseHistory.js
```diff
@@ -0,0 +1,197 @@
...
+        lease.plot_searches.forEach((plotSearch) => {
+          historyItems.push({
+            key: `plot-search-${plotSearch.name}-${Math.random().toString()}`,
+            id: plotSearch.id,
+            itemTitle: plotSearch.name,
+            startDate: plotSearch.begin_at,
+            endDate: plotSearch.end_at,
+            plotSearchType: plotSearch.type,
+            plotSearchSubtype: plotSearch.subtype,
+            itemType: "Haku",
```
henrinie-nc, line 80: Is this string used internally in the app only and not shown to the users? If shown, would it need to be translated?

src/leases/components/leaseSections/summary/LeaseHistory.js
```diff
@@ -0,0 +1,197 @@
...
+              id: content_object.id,
+              itemTitle: content_object.name,
+              startDate: content_object.begin_at,
+              endDate: content_object.end_at,
+              plotSearchType: content_object.type,
+              plotSearchSubtype: content_object.subtype,
+              itemType: "Haku",
+            })
+          }
+          if (relatedPlotApplication.content_type?.model === "targetstatus") {
```
henrinie-nc, line 125: Would it make sense to make these enums? ContetType model names

src/leases/components/leaseSections/summary/LeaseHistory.js
```diff
@@ -0,0 +1,197 @@
...
+              itemTitle: content_object.identifier,
+              applicantName: `${content_object.applicant_names.join(" ")}`,
+              receivedAt: content_object.received_date,
+              itemType: "Aluehakemus",
+            })
+          }
+        })
+      }
+
+      let leaseProps: any = {
```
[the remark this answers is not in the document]
henrinie-nc: Could this be `const` instead?

src/leases/components/leaseSections/summary/LeaseHistoryEdit.js
```diff
@@ -104,17 +108,29 @@ class RelatedLeasesEdit extends Component<Props, State> {
...
     this.setState({
-      newLease: toLease,
+      newLease: newHistoryItem,
     });
 
-    createReleatedLease({
-      from_lease: currentLease.id,
-      to_lease: toLease.value,
-    });
+    switch (newHistoryItem.type) {
```
henrinie-nc, line 118: Could also be if since only two conditions

src/leases/components/leaseSections/summary/LeaseHistoryEdit.js
```diff
@@ -156,12 +178,122 @@ class RelatedLeasesEdit extends Component<Props, State> {
...
+        lease.plot_searches.forEach((plotSearch) => {
+          historyItems.push({
+            key: `plot-search-${plotSearch.name}-${Math.random().toString()}`,
+            id: plotSearch.id,
+            itemTitle: plotSearch.name,
+            startDate: plotSearch.begin_at,
+            endDate: plotSearch.end_at,
+            plotSearchType: plotSearch.type,
+            plotSearchSubtype: plotSearch.subtype,
+            itemType: "Haku",
```
henrinie-nc, line 203: This `itemType` was also used in another file. I wondered whether it is shown to the user. But since it is here again, how about making it enum?

src/leases/components/leaseSections/summary/LeaseHistoryEdit.js
```diff
@@ -156,12 +178,122 @@ class RelatedLeasesEdit extends Component<Props, State> {
...
+        id: lease.id,
+        deleteId: lease.related_lease_id,
+        lease: lease,
+        startDate: lease.start_date,
+        endDate: lease.end_date,
+        onDelete: handleDelete
+      }
+
+      // active will be a number when used in a map function
+      if (typeof active === "boolean") {
```
henrinie-nc, line 289: What is this used for?

src/leases/components/leaseSections/summary/LeaseHistoryItem.js
```diff
@@ -22,24 +22,55 @@ import type {UsersPermissions as UsersPermissionsType} from '$src/usersPermissio
...
+    itemType = '',
+    onDelete,
+    stateOptions,
+    usersPermissions
+  }: Props,
+) => {
+  const titleString = lease ? getContentLeaseIdentifier(lease) : itemTitle;
+  const MAX_TITLE_LENGTH = 16
+  const title = getTitleText(titleString, MAX_TITLE_LENGTH)
+  const externalLinkHref = lease ? `${getRouteById(Routes.LEASES)}/${lease.id}`
```
henrinie-nc, line 66: This is a pretty long ternary chain

src/leases/components/leaseSections/summary/LeaseHistoryItem.js
```diff
@@ -22,24 +22,55 @@ import type {UsersPermissions as UsersPermissionsType} from '$src/usersPermissio
...
+    : itemType === 'Haku' && id ? `${getRouteById(Routes.PLOT_SEARCH)}/${id}`
+    : itemType === 'Hakemus' && id ? `${getRouteById(Routes.PLOT_APPLICATIONS)}/${id}`
+    : itemType === 'Aluehakemus' && id ? `${getRouteById(Routes.AREA_SEARCH)}/${id}`
+    : null
+  // TODO: Add permissions for deleting plot searches and plot applications
```
henrinie-nc, line 71: What are these TODO's here?

src/leases/enums.js
```diff
@@ -109,6 +109,7 @@ export const LeaseState = {
   RYA: 'rya',
   APPLICATION: 'application',
   TRANSFERRED: 'transferred',
+  PLOTSEARCH: 'plotsearch',
```
henrinie-nc, line 112: Why was this added, is it used in this new code?

src/leases/helpers.js
```diff
@@ -150,6 +150,22 @@ export const getFullAddress = (item: Object): ?string => {
...
+ * Get a lease history item title text that is truncated when it is long
+ * @param {?string} text
+ * @param {number} MAX_TITLE_LENGTH
+ * @returns {string}
+ */
+
+export const getTitleText = (text: ?string, maxLength: number) => {
+  if (text) {
+    return text.length > maxLength ? text.substr(0, maxLength) + "..." : text
+  } else {
```
henrinie-nc, line 163: Wondering the purpose of this else, could the whole if-else be removed and use `return text.length > maxLength ? text.substr(0, maxLength) + "..." : text`

src/leases/helpers.js
```diff
@@ -3118,3 +3134,39 @@ export const clearUnsavedChanges = () => {
 export const getLeasesWithContractNumber = (leasesForContractNumbers: LeaseList): boolean => {
   return (get(leasesForContractNumbers, 'count') > 0);
 };
+
+/**
+ * Destructures a nested lease object in a related lease.
+ * @param {Object} lease
```
henrinie-nc, line 3140: In this typing system is it possible to create a type for lease and use it here?

src/leases/helpers.js
```diff
@@ -3118,3 +3134,39 @@ export const clearUnsavedChanges = () => {
...
+ * @returns {Object}
+ */
+export const sortRelatedHistoryItems = (a: Object, b: Object): Object => {
+  let aTime = a.startDate || a.receivedAt || null
+  let bTime = b.startDate || b.receivedAt || null
+
+  if (aTime && bTime && aTime != bTime) {
+
+    if (aTime < bTime) {
```
henrinie-nc, line 3163: Might be simpler to write
```
if (aTime > bTime) return 1
if (aTime < bTime) return -1
return 0
```

henrinie-nc:
Looks good overall, pointed out some small improvements. Would be a good idea to format the files so the code style would be similar

src/components/inputs/LeaseSelectInput.js
EmiliaMakelaVincit: Something I noticed from the emails was this point which is moot, as the preparer UI is only in Finnish and does not have translations for its strings. (It is also hard coded to add an Accept-Language header to all API requests to retrieve data in translatable Django models in Finnish whenever possible.)

src/components/inputs/LeaseSelectInput.js
NC-jsAhonen: The label is the label of the option of the select element which this function populates with the fetched results.

The slice is for limiting the number of items in the dropdown, but I just realised that the CSS limits the height of the dropdown, and makes it scrollable.

src/leases/components/leaseSections/summary/LeaseHistoryItem.js
NC-jsAhonen: Actually, what needs to be done is to rename permissions `delete_relatedlease` and `add_relatedlease`.

henrinie-nc: Will you be able to do that and remove the TODO's?

src/leases/helpers.js
NC-jsAhonen: That ternary won't do, because if text is undefined, then the function would return a null. But I want this function to always return a string, empty or not. And I added the if-else statement so that I wouldn't need to include the null check in the ternary statement.

src/components/inputs/LeaseSelectInput.js
henrinie-nc: Would like to receive insights on this question, why not implement with Promise.all?

henrinie-nc (approved):
Most things seem to have been fixed. I reposted my question about using Promise.all in `getHistoryItems` to get the response faster, would like to hear your thoughts on that.

src/components/inputs/LeaseSelectInput.js
NC-jsAhonen: In brief, it would lead to unnecessary complications in the code. As far as I understand the logic, Promise.all() would create a single list of results that would then passed into the getHistoryItemOptions callback function. But if we do that, we would need to sort out the types of the four different kinds of data anyway in order to get the correct values for the label, type and content_type_model. It is easier to sort these out when the result sets from each endpoint is kept separate and then later combined into a single list.

henrinie-nc: Promise.all() would not create a single list of result, it would return each Promise individually. The benefit would be that the fetches are run simultaneously and would possibly take less time. How I see the current code is that it does one fetch, waits for response, does another fetch, waits, and so on.

I suggested it purely to make all the calls at once and reduce the time waiting for the responses one by one.

NC-jsAhonen: Done

PR #449, opened 2023-11-21
Fix plot search list page filtering

EmiliaMakelaVincit:
- Change redux-form names to match the backend
- Reflect the above in simple mode detection on pageload
- Add proper subtype handling (only show subtypes of selected type, clear subtype on type change, disable dropdown with no type selected)

Matching backend: https://github.com/City-of-Helsinki/mvj/pull/556

### DIFF
--- a/src/plotSearch/components/PlotSearchListPage.js
+++ b/src/plotSearch/components/PlotSearchListPage.js
@@ -73,9 +73,9 @@ const visualizationTypeOptions = [
   {value: VisualizationTypes.MAP, label: 'Kartta', icon: <MapIcon className='icon-medium' />},
 ];
 
-type OwnProps = {
+type OwnProps = {|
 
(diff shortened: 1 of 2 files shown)


src/plotSearch/components/PlotSearchListPage.js
```diff
@@ -73,9 +73,9 @@ const visualizationTypeOptions = [
   {value: VisualizationTypes.MAP, label: 'Kartta', icon: <MapIcon className='icon-medium' />},
 ];
 
-type OwnProps = {
+type OwnProps = {|
 
-};
+|};
```
PasiVuohijoki, line 78: I needed to google what the pipes are doing here. But yeah, makes sense now when I know :D

PasiVuohijoki (approved):

PR #450, opened 2023-12-04
Fix getting targetIdentifier if plan_unit falsy

henrinie-nc:
Code seems to intend to pick `target.custom_detailed_plan` if `target.plan_unit` does not exist. The code failed on `target.plan_unit` being falsy, so couldn't get the identifier.


antti-honkanen (approved):
👍

PR #451, opened 2023-12-22
Fix: stringify form field default_value for multiple selection feature

henrinie-nc:
API expects `default_value` to be of type string.
This ensures that all values are stringified.

### DIFF
--- a/src/plotSearch/components/plotSearchSections/application/EditPlotApplicationSectionFieldForm.js
+++ b/src/plotSearch/components/plotSearchSections/application/EditPlotApplicationSectionFieldForm.js
@@ -398,26 +398,20 @@ const EditPlotApplicationSectionFieldForm = ({
       change(
         FormNames.PLOT_SEARCH_APPLICATION_SECTION_STAGING,
         `${field}.default_value`,
-        JSON.stringify( // default_value must be a string
-          fieldValues.default_value?.[0] || ''
-        ),
+        fieldValues.default_value?.[0] || ''
       );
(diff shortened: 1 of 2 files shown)


henrinie-nc:
Not the correct place to implement the fix, need to move it to the helper so that it happens before sending the request.

EmiliaMakelaVincit:
I've added my suggestion to fix this properly and it did seem to save properly for me and also work both when opening that section for further editing and in the public UI.

I did note a few tangentially related issues when saving form changes that might've been overlooked when the editor was first implemented. Namely:
- Adding a new subsection doesn't seem to work at all: `forms.models.form.Section.form.RelatedObjectDoesNotExist: Section has no form.` The `form_id` field is submitted for pre-existing sections and could be manually inserted for new sections, but the backend could just fill this in on its own upon section creation too.
- Adding a field that uses choices and also specifying any choices at the same time fails: `ERROR:  null value in column "field_id" violates not-null constraint`. The backend should insert the ID of the newly created field here.

henrinie-nc:
I have tested the changes locally and it seems to fix the original issue.

Regarding the reported new issues:

* Adding a new subsection indeed fails
    * I'm not entirely sure what is missing, but there seems to be more than just the `form_id`, e.g. `id`, `parent_id`, and for the `fields` it is also missing `id`. Would need deeper investigation to the issue, I will create a ticket.
* Adding field with prefilled choices indeed fails
    * Saving it first without the choices works as a workaround

PR #456, opened 2024-02-01
Chore/update node gh actions

henrinie-nc:
Partly testing whether the app just builds with node18

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -9,7 +9,7 @@ jobs:
   sonarcloud:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
       with:
         fetch-depth: 0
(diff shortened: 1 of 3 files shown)


antti-honkanen:
Changes look valid, but pipeline fails.

henrinie-nc:
> Changes look valid, but pipeline fails.

Indeed, was testing what happens. Dependencies need updates to be able to build with node18, working on that.

antti-honkanen:
Now that the build works, should we remove commented code?

henrinie-nc:
> Now that the build works, should we remove commented code?

We should remove commented code, I was hesitant because I didn't get them to work, and wasn't sure if they might reveal some issues later (would be easier to maybe see these were here).
For instance the favicon generator seems to generate all kinds of icons for mobile, but that is not needed afaik.

But I will remove them since that's how it should be done

antti-honkanen (approved):

PR #462, opened 2024-02-26
Fix fieldtype mapping after removal of FieldType in backend

henrinie-nc:
Depends on https://github.com/City-of-Helsinki/mvj/pull/592


antti-honkanen (approved):
Looks good.

PR #471, opened 2024-04-03
Fix field types to render area search application form

NC-jsAhonen:
This fixes the issue of area search application form not rendering properly in the mvj-ui, when creating an area search.


henrinie-nc (approved):
lgtm

PR #473, opened 2024-04-16
Add dev option to compile script

NC-jsAhonen:
`yarn run compile` will create a production build.

Builds for the dev environment are created manually with this script too, but they need to have NODE_ENV as 'development'.

This change introduces the --dev flag to the script, so that running `yarn run compile --dev` will create a development build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,13 @@ yarn start
 ```bash
 yarn run compile
 ```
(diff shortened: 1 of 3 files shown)


NC-jsAhonen:
This doesn't work yet. Closing

PR #474, opened 2024-04-19
Area search application: add and remove attachments in officer UI

NC-jsAhonen:
Enables adding and removing of attachments to and from area search applications by the Helsinki city official.

### DIFF
--- a/src/areaSearch/actions.js
+++ b/src/areaSearch/actions.js
@@ -41,6 +41,8 @@ import type {
   UploadAreaSearchAttachmentAction,
   ReceiveFileOperationFinishedAction,
   ReceiveFileOperationFailedAction,
+  SetAreaSearchAttachmentsAction,
+  UploadedAreaSearchAttachmentMeta,
 } from '$src/areaSearch/types';
 import type {Attributes, Methods} from '$src/types';
 
@@ -160,3 +162,6 @@ export const receiveFileOperationFinished = (): ReceiveFileOperationFinishedActi
 
 export const receiveFileOperationFailed = (error: any): ReceiveFileOperationFailedAction =>
   createAction('mvj/areaSearch/RECEIVE_FILE_OPERATION_FAILED')(error);
+
+export const setAreaSearchAttachments = (attachments: Array<UploadedAreaSearchAttachmentMeta>): SetAreaSearchAttachmentsAction =>
+  createAction('mvj/areaSearch/SET_ATTACHMENTS')(attachments);
--- a/src/areaSearch/components/AreaSearchApplication.js
+++ b/src/areaSearch/components/AreaSearchApplication.js
@@ -18,7 +18,10 @@ import {
 } from '$util/helpers';
 import type {Attributes} from '$src/types';
 import {reshapeSavedApplicationObject} from '$src/plotApplications/helpers';
-import {transformApplicantSectionTitle} from '$src/application/helpers';
+import {
+  transformApplicantSectionTitle,
+  getAreaSearchApplicationAttachmentDownloadLink,
+} from '$src/application/helpers';
 import Title from '$components/content/Title';
 import Divider from '$components/content/Divider';
 import Collapse from '$components/collapse/Collapse';
@@ -212,15 +215,16 @@ class AreaSearchApplication extends Component<Props, State> {
             ]}
           />)}
         <Collapse headerTitle="Liitteet" defaultOpen>
-          {areaSearch.area_search_attachments.map((file, index) => <Row key={file.id}>
-            <Column small={3}>Liite {index + 1}</Column>
-            <Column small={9}>
-              <FileDownloadLink
-                fileUrl={file.file}
-                label={file.name}
-              />
-            </Column>
-          </Row>)}
+          {areaSearch.area_search_attachments.map((file, index) => {
+            return (<Row key={file.id}>
+              <Column small={3}>Liite {index + 1}</Column>
+              <Column small={9}>
+                <FileDownloadLink
+                  fileUrl={getAreaSearchApplicationAttachmentDownloadLink(file.id)}
+                  label={file.name}
+                />
+              </Column>
+            </Row>)})}
           {areaSearch.area_search_attachments.length === 0 && <p>
             Hakemuksella ei ole liitteitä.
           </p>}
--- a/src/areaSearch/components/AreaSearchApplicationEdit.js
+++ b/src/areaSearch/components/AreaSearchApplicationEdit.js
@@ -1,5 +1,4 @@
 // @flow
-
 import React, {Component, Fragment} from 'react';
 import {connect} from 'react-redux';
 import flowRight from 'lodash/flowRight';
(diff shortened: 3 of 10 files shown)


src/areaSearch/components/AreaSearchApplicationEdit.js
```diff
@@ -91,14 +96,56 @@ class AreaSearchApplicationEdit extends Component<Props, State> {
 
     initialize(getInitialAreaSearchEditForm(areaSearch));
   }
+  
+  handleFileAdded = (e: any) => {
```
henrinie-nc, line 100: Can `e` be anything?

src/areaSearch/actions.js
```diff
@@ -43,6 +43,7 @@ import type {
   ReceiveFileOperationFailedAction,
 } from '$src/areaSearch/types';
 import type {Attributes, Methods} from '$src/types';
+import type { SetAreaSearchAttachmentsAction } from "./types";
```
henrinie-nc, line 46: `'` or `"`? I think we should use one, not both. This seems to apply in multiple files, please check

src/areaSearch/components/AreaSearchApplication.js
```diff
@@ -40,6 +40,10 @@ import AreaSearchStatusNoteHistory from '$src/areaSearch/components/AreaSearchSt
 import {getFormAttributes, getIsFetchingFormAttributes} from '$src/application/selectors';
 import {APPLICANT_SECTION_IDENTIFIER} from '$src/application/constants';
 import type {Form} from '$src/application/types';
+import {
+  getApplicationAttachmentDownloadLink,
```
henrinie-nc, line 44: Is this used in this file?

src/areaSearch/components/AreaSearchApplicationEdit.js
```diff
@@ -41,21 +41,26 @@ import {getInitialAreaSearchEditForm, transformApplicantInfoCheckTitle} from '$s
...
+import RemoveButton from '$components/form/RemoveButton';
+import type { UploadedAreaSearchAttachmentMeta } from "../types";
 
-type OwnProps = {
-
-};
 
 type Props = {
-  ...OwnProps,
   areaSearch: Object | null,
```
henrinie-nc, line 55: Does `areaSearch` have a type?

src/areaSearch/components/AreaSearchApplicationEdit.js
```diff
@@ -91,14 +96,56 @@ class AreaSearchApplicationEdit extends Component<Props, State> {
...
+      },
+    });
+  }
+  
+  handleDeleteFile = (file: File) => {
+    const {deleteUploadedAttachment, areaSearch, setAreaSearchAttachments} = this.props;
+    const currentFiles = areaSearch?.area_search_attachments || [];
+    
+    if (file.id) {
+      deleteUploadedAttachment({
```
henrinie-nc, line 119: It was a requirement that files can not be deleted

src/areaSearch/components/AreaSearchApplicationEdit.js
```diff
@@ -249,18 +296,28 @@ class AreaSearchApplicationEdit extends Component<Props, State> {
...
+            <Column small={8}>
               <FileDownloadLink
-                fileUrl={file.file}
+                fileUrl={getAreaSearchApplicationAttachmentDownloadLink(file.id)}
                 label={file.name}
               />
             </Column>
-          </Row>)}
+            <Column small={1}>
+              <RemoveButton onClick={() => this.handleDeleteFile(file)} />
```
henrinie-nc, line 309: It was a requirement that files can't be deleted, please check that

src/areaSearch/components/AreaSearchApplicationPage.js
```diff
@@ -147,26 +148,26 @@ class AreaSearchApplicationPage extends Component<Props, State> {
       fetchFormAttributes,
       fetchApplicantInfoCheckAttributes,
     } = this.props;
-
+
```
henrinie-nc, line 151: Check the added whitespace in this file

src/areaSearch/reducer.js
```diff
@@ -13,6 +13,8 @@ import type {
 import type {ReceiveAttributesAction, ReceiveMethodsAction} from '$src/areaSearch/types';
 import type {ApiResponse} from '$src/types';
 import merge from 'lodash/merge';
+import type { Set } from "immutable";
```
henrinie-nc, line 16: This also has `' and "` plus `{ thing } and {thing}`

src/areaSearch/reducer.js
```diff
@@ -82,6 +84,13 @@ const currentAreaSearchReducer: Reducer<Object | null> = handleActions({
   },
   ['mvj/areaSearch/CREATE_SPECS']: () => null,
   ['mvj/areaSearch/RECEIVE_SPECS_CREATED']: (state, {payload: result}) => result,
+  ['mvj/areaSearch/SET_ATTACHMENTS']: (state, {payload}) => {
```
henrinie-nc, line 87: does `payload` have any type?

henrinie-nc (changes requested):
Looks like this would work, check the comments for some fine tuning. If typing could be improved that would be fantastic!

Also the feature had a requirement that files can't be deleted, that should probably be applied here.

src/areaSearch/components/AreaSearchApplicationEdit.js
NC-jsAhonen: Looking at the codebase, I think it is Event.

src/areaSearch/components/AreaSearchApplicationEdit.js
```diff
@@ -249,18 +296,28 @@ class AreaSearchApplicationEdit extends Component<Props, State> {
...
+            </Column>
+          </Row>)})}
           {areaSearch.area_search_attachments.length === 0 && <p>
             Hakemuksella ei ole liitteitä.
           </p>}
+          <AddFileButton
+            label='Lisää tiedosto'
+            onChange={this.handleFileAdded}
+            name="AreaSearchApplicationEditAttachment"
+            disabled={isPerformingFileOperation}
```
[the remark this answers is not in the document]
NC-jsAhonen: I can't test this, because my browser does not support selecting multiple files at once.

src/areaSearch/reducer.js
NC-jsAhonen: This was also an unused module, so I removed it.

src/areaSearch/reducer.js
NC-jsAhonen: I defined it, and renamed it as attachments, because it does not create an HTTP request but only updates the state locally.

src/areaSearch/components/AreaSearchApplicationEdit.js
```diff
@@ -91,14 +97,39 @@ class AreaSearchApplicationEdit extends Component<Props, State> {
 
     initialize(getInitialAreaSearchEditForm(areaSearch));
   }
+  
+  handleFileAdded = (e: Event) => {
+    const {uploadAttachment, areaSearch, setAreaSearchAttachments} = this.props;
+    const currentFiles = areaSearch?.area_search_attachments || [];
+    const file = e.target.files[0];
```
henrinie-nc, line 104: Does this support adding multiple files at once, or is it going to run once for each file?

src/areaSearch/components/AreaSearchApplicationEdit.js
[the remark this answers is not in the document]
henrinie-nc: In that case the fileinput probably allows one file at a time, which is ok

src/areaSearch/types.js
```diff
@@ -27,9 +27,42 @@ export type AreaSearchState = {
   isPerformingFileOperation: boolean,
 };
 
-export type AreaSearch = Object;
+export type AreaSearch = {
```
henrinie-nc, line 30: :clap:

henrinie-nc (approved):
Nicely added typing, good to go

PR #477, opened 2024-05-03
move service unit filter next to basic search in leases view

robertrytovuori:
Screenshot-service-unit-filter

### DIFF
--- a/src/components/table/TableFilters.js
+++ b/src/components/table/TableFilters.js
@@ -10,7 +10,7 @@ type Props = {
   filterOptions: Array<Object>,
   filterValue: Array<string>,
   onFilterChange?: Function,
-  componentToRenderUnderLabel?: React$Node,
+  componentToRenderUnderTitle?: React$Node,
 }
 const TableFilters = ({
   alignFiltersRight,
@@ -18,7 +18,7 @@ const TableFilters = ({
   filterOptions,
   filterValue,
   onFilterChange,
-  componentToRenderUnderLabel = null
+  componentToRenderUnderTitle = null
 }: Props): React$Node => {
   const handleFilterChange = (values: Array<Object>) => {
     if(onFilterChange) {
@@ -35,7 +35,7 @@ const TableFilters = ({
           {!!filterOptions.length &&
             <p className='table__filters_filter-wrapper_title'>Suodatus</p>
           }
-          {!!componentToRenderUnderLabel && componentToRenderUnderLabel}
+          {!!componentToRenderUnderTitle && componentToRenderUnderTitle}
           <CheckboxInput
             checkboxName='lease-type-checkbox'
             legend='Suodata'
--- a/src/components/table/TableFilters.js
+++ b/src/components/table/TableFilters.js
@@ -10,7 +10,7 @@ type Props = {
   filterOptions: Array<Object>,
   filterValue: Array<string>,
   onFilterChange?: Function,
-  componentToRenderNextToLabel?: React$Node,
+  componentToRenderUnderLabel?: React$Node,
 }
 const TableFilters = ({
   alignFiltersRight,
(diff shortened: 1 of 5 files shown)


src/leases/components/search/Search.js
```diff
@@ -236,6 +236,27 @@ class Search extends PureComponent<Props, State> {
...
+        <SearchRow>
+          <SearchLabelColumn>
+            <SearchLabel>Palvelukokonaisuus</SearchLabel>
+          </SearchLabelColumn>
+          <SearchInputColumn>
+            <FormField
+              autoBlur
+              disableDirty
+              fieldAttributes={{
+                label: 'Palvelukokonaisuus',
```
NC-jsAhonen, line 248: This label is repeated in many places, so it is recommended to use the enum in `leases/enums.js`

src/leases/components/search/Search.js
```diff
@@ -236,6 +236,27 @@ class Search extends PureComponent<Props, State> {
             />
           </Column>
         </Row>
+        <SearchRow>
+          <SearchLabelColumn>
+            <SearchLabel>Palvelukokonaisuus</SearchLabel>
```
NC-jsAhonen, line 241: Could this be something like `Suodata palvelukokonaisuuden mukaan`, to distinguish from the top nav bar filter. Should this label be put on a separate line above the field, just like in other fields?

You could also add a visible label for the search field above in the same way with text something like `Hae hakusanalla` or something else.

NC-jsAhonen (changes requested):
The functionality works as expected. I added some suggestions for improving layout and following established practices.

src/leases/components/search/Search.js
lebanonsky: Headerissa voisi lukea **Oma palvelukokonaisuus**
Suodatuksessa voisi lukea **Vuokrauspalvelu**

robertrytovuori:
Screenshot-service-unit-filter

src/leases/components/search/Search.js
```diff
@@ -105,7 +105,7 @@ class Search extends PureComponent<Props, State> {
       newState.municipalityOptions = getFieldOptions(props.leaseAttributes, LeaseFieldPaths.MUNICIPALITY);
       newState.tenantTypeOptions = getFieldOptions(props.leaseAttributes, LeaseTenantContactSetFieldPaths.TYPE, false);
       newState.typeOptions = getFieldOptions(props.leaseAttributes, LeaseFieldPaths.TYPE);
-      newState.serviceUnitOptions = getFieldOptions(props.leaseAttributes, 'service_unit', true);
+      //newState.serviceUnitOptions = getFieldOptions(props.leaseAttributes, 'service_unit', true);
```
NC-jsAhonen, line 108: `isSearchBasicMode` method beginning on line 147 `delete searchQuery.service_unit;` needs to be added so that the detailed search view would not render when the service unit filter is being modified. Now it will render when the page is refreshed.

NC-jsAhonen (changes requested):
Posted some new comments

src/components/table/TableFilters.js
```diff
@@ -33,6 +35,7 @@ const TableFilters = ({
           {!!filterOptions.length &&
             <p className='table__filters_filter-wrapper_title'>Suodatus</p>
           }
+          {!!componentToRenderUnderLabel && componentToRenderUnderLabel}
```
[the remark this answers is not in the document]
robertrytovuori: Renamed it to componentToRenderUnderTitle. Though serviceUnitFilterComponent tells more in this use case, I still think the TableFilter component does not have to know what React component is being injected.

src/leases/components/search/Search.js
robertrytovuori: fixed

NC-jsAhonen (approved):
Approved

PR #479, opened 2024-05-08
fix checking changes in subventions

NC-jsAhonen:
The "Määrä" was not updating properly when editing rent adjustments in the Vuokraukset => Vuokrat tab. Here are the fixes.


robertrytovuori (approved):
Looks good

PR #480, opened 2024-05-15
Mvj 248 and Mvj 216 missing icons

robertrytovuori:
The issue with icons not showing was that we have moved to Webpack 5 and file loading is handled a bit differently in the new Webpack version

### DIFF
--- a/config/webpack.config.dev.js
+++ b/config/webpack.config.dev.js
@@ -90,7 +90,10 @@ module.exports = {
       },
       {
         test: /\.(eot|svg|png|jpe?g)?(\?v=[0-9]\.[0-9]\.[0-9])?$/,
-        loader: require.resolve('file-loader'),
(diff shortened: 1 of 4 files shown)


henrinie-nc (changes requested):
Is `'file-loader` package used any longer?

robertrytovuori:
I get an error from url-loader if file-loader dependency isn't there:
```
Module build failed (from ./node_modules/url-loader/dist/cjs.js):
Error: Cannot find module 'file-loader
```

henrinie-nc:
> I get an error from url-loader if file-loader dependency isn't there:
> 
> ```
> Module build failed (from ./node_modules/url-loader/dist/cjs.js):
> Error: Cannot find module 'file-loader
> ```

Then this is good to go

henrinie-nc (approved):

PR #481, opened 2024-05-23
Develop to master

henrinie-nc:


NC-jsAhonen (approved):

NC-jsAhonen (changes requested):
Are the tonttihaku and aluehaku hidden, and should they be?

henrinie-nc:
> Are the tonttihaku and aluehaku hidden, and should they be?

@NC-jsAhonen 
For production should be hidden, see https://github.com/City-of-Helsinki/mvj-ui/pull/455

NC-jsAhonen (approved):
The SideMenu seems to have the hiding of the area and plot searches for production, so it's okay. Otherwise, it doesn't seem to me that any further fixes need to be done at this point.

PR #483, opened 2024-05-24
Related lease in create lease: remove service unit filter

NC-jsAhonen:
Remove lease filter from Luo vuokraustunnus modal field "Liittyy vuokraukseen".

### DIFF
--- a/src/components/modal/_modal.scss
+++ b/src/components/modal/_modal.scss
@@ -10,7 +10,7 @@
     left: 50%;
     transform: translateX(-50%) translateY(-50%);
     box-shadow: 0 0 5px 0 rgba(0, 0, 0, 0.06);
-    overflow: auto;
+    overflow: visible;
 
     @include transition(0.5s, top, ease);
   }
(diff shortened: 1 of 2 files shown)


src/components/modal/_modal.scss
```diff
@@ -10,7 +10,7 @@
     left: 50%;
     transform: translateX(-50%) translateY(-50%);
     box-shadow: 0 0 5px 0 rgba(0, 0, 0, 0.06);
-    overflow: auto;
+    overflow: visible;
```
NC-jsAhonen, line 13: image

NC-jsAhonen (commented):
Added an illustration

src/components/modal/_modal.scss
NC-jsAhonen: This change makes it easier to go through the search results. The overflow prop applies to all modals, but I haven't come up with any problem that might come from this.

robertrytovuori (approved):

PR #484, opened 2024-05-27
toggleable usergroup list

NC-jsAhonen:
The new usergroup list looks as follows. It is toggled using the info icon. I could not find a person icon and the icons are created in the mvj-ui as manually inserted vector values, so it would take extra work to make it into a person icon. Is info icon good enough in your opinion?

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "file-saver": "^2.0.5",
     "font-awesome": "^4.7.0",
     "foundation-sites": "6.5.3",
+    "hds-react": "^3.8.0",
     "history": "^4.9.0",
     "leaflet": "^1.5.1",
     "leaflet-draw": "^1.0.4",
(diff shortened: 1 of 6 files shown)


robertrytovuori:
Have you tried looking online for free person icons? I found for example this: https://www.svgrepo.com/svg/506667/person

NC-jsAhonen:
> Have you tried looking online for free person icons? I found for example this: https://www.svgrepo.com/svg/506667/person

No, because I am also wondering whether info icon is a better idea that a person icon, given that the only thing it does is give some info. It is not a regular user account dropdown menu, so I am a bit doubtful whether info or person icon is better.

henrinie-nc:
Should we use HDS icons in case we migrate to use HDS at some stage?
https://hds.hel.fi/foundation/visual-assets/icons/list/

robertrytovuori (approved):

NC-jsAhonen:
image

PR #485, opened 2024-05-28
Mvj 249 flow to typescript conversion

robertrytovuori:
This is a huge PR. Not sure what is the best approach to reviewing it since pretty much every .js and .jsx file has been transformed to .ts and .tsx. Also had to make some changes in webpack and other places to make the typescript work. I have ran the tests, compiled and ran the dev server.


package.json
```diff
@@ -32,6 +32,7 @@
     "copy-to-clipboard": "^3.2.0",
     "date-fns": "^2.0.0-beta.4",
     "dotenv": "^8.0.0",
+    "file-loader": "^6.2.0",
```
henrinie-nc, line 35: It sounds a bit like `file-loader`, `ts-loader`, and `utility-types` should be devDependencies

henrinie-nc (approved):
Big chunk of changes to review. I looked a bit here and there, nothing significant popped up.
One thing I noted was addition of some dependencies, which seemed like development dependencies, please check those and move to devDependencies if they belong there. Approving ahead of time

PR #486, opened 2024-05-31
change type for formatDecimalNumber to take in string

henrinie-nc:
Please check if this makes sense, I encoutered errors on `vuokra-alue` page due to how this type was previously. I'm not sure how this changed, or if it was wrong to begin with


robertrytovuori:
I must have mistakenly altered the code to match the typing in this case when I should have done it the other way around. Good find

robertrytovuori (approved):

PR #495, opened 2024-06-26
fix typescript errors

robertrytovuori:


src/creditDecision/types.ts
```diff
@@ -11,9 +11,9 @@ export type CreditDecisionState = {
   isFetchingHistoryByBusinessId: IsFetchingHistoryMap;
   isFetchingHistoryByContactId: IsFetchingHistoryMap;
 };
-export type HistoryDataMap = Record<string, Record<string, any>>;
+export type HistoryDataMap = any;
```
NC-jsAhonen, line 14: Can you come up with some more specific typing for these?

NC-jsAhonen (changes requested):
It is recommended to not use any types

PR #498, opened 2024-06-27
Ahjo decision links

NC-jsAhonen:
The current links to Ahjo decisions will be deprecated. These are the new ones.


robertrytovuori (approved):
Looks good

PR #501, opened 2024-07-01
MVJ-297 Typescript error top navigation

robertrytovuori:


NC-jsAhonen (changes requested):
Just remove the TopNavigationSettings import and then it seems good to go.

PR #507, opened 2024-07-30
Mvj 277 typing fixes

henrinie-nc:
Adds some type declarations to fix typescript warnings, fix some typing, adjust tsconfig


NC-jsAhonen (approved):
Looks ok

PR #509, opened 2024-08-08
MVJ-394 MVJ-UI refactor area search into hook syntax

NC-jsAhonen:
These changes migrate the code from the older and less documented class syntax into the current standard React practice of hooks.

ATTENTION
There was a previously existing bug that is unrelated to this change: the attachments are not rendered immediately after they are added to area search application forms, and the UI gets stuck after doing so.


henrinie-nc:
@NC-jsAhonen Is this now ready for review?

PR #511, opened 2024-08-14
Properly split the query parameter for service unit in email reports

juho-kettunen-nc:
Service unit query parameter was previously sent as single string like "1,2,3,4", which is not a proper input for forms.ModelMultipleChoiceField. Synchronous reports already handled this query string properly, so now email reports use the same query parameter parsing for service unit.


henrinie-nc (approved):
Looks good and seems to fix the issue

PR #512, opened 2024-08-15
fix icons in send email modal

robertrytovuori:
I couldn't get the awesome-font icons to load in the send email modal. It should work, I just don't understand why not. Therefore, I made a bit of a different solution by using text instead of icons but I think it looks quite good like this as well.

send-email-modal-screenshot


henrinie-nc (approved):
Well that is a great use of text as icons :)

henrinie-nc:
@robertrytovuori One question or improvement idea, is it possible to have a tooltip / is there one for these buttons? That would explain what the button does

robertrytovuori:
There is a tooltip on hover that explains the buttons

PR #517, opened 2024-08-29
Remove automatically generated invoices not editable text and restore edit buttons

NC-jsAhonen:
Fields of automatically generated invoices are mostly read only, but there are exceptions: payments and postpone_date.

For this reason, the warning text will be removed and the save button restored for generated invoices.


henrinie-nc (approved):
lgtm

PR #521, opened 2024-09-13
Mvj 453 hds login

henrinie-nc:
I would like feedback on:
* are the auth states being handled correctly, is this missing some situations that might occur?
* sanity check the tests, is it testing what it seems or not?

Also if you see something obvious that triggers rerenders that would be nice. Not really ideal that e.g. `USER_FOUND` actions trigger multiple times. Is there a better way to define the `useEffect`'s (One cause is probably `mapStatesToPros` on `App`, could see if something can be fixed there.

Check the docs here: https://hds.hel.fi/components/login/


src/auth/components/CallbackPage.tsx
```diff
@@ -1,33 +1,28 @@
...
-
-class CallbackPage extends PureComponent<Props> {
-  successCallback = () => {
-    const {
-      history
-    } = this.props;
+const CallbackPage = (props: Props) => {
+  const onSuccess = (user: User) => {
+    const { history } = props;
     history.push(getRedirectUrlFromSessionStorage() || getRouteById(Routes.LEASES));
```
henrinie-nc, line 15: Check if the url startswith `/callback` and don't store that into session storage. Could happen if auth callback from oidc server fails.

src/auth/useAuth.ts
```diff
@@ -0,0 +1,77 @@
...
+const useAuth = () => {
+  const [loggedIn, setLoggedIn] = useState(false);
+  const { login: oidcLogin, logout: oidcLogout, isRenewing: oidcIsRenewing } = useOidcClient();
+  const authenticatedUser = useAuthenticatedUser();
+  const dispatch = useDispatch();
+  const [apiTokensClientSignal, apiTokensClientSignalReset] = useApiTokensClientTracking();
+  const { getStoredApiTokens } = useApiTokens();
+
+  const setLoggedInIfApiTokenExists = useCallback(() => {
+    const [_error, apiToken] = getStoredApiTokens();
```
henrinie-nc, line 26: Consider logging possible error

src/auth/actions.ts
```diff
@@ -1,6 +1,10 @@
-import { createAction } from "redux-actions";
-import type { ClearApiTokenAction, FetchApiTokenAction, ReceiveApiTokenAction, TokenNotFoundAction } from "./types";
+import { createAction } from 'redux-actions';
```
juho-kettunen-nc, line 1: Why does this file use single quotes while most other files use double quotes?

juho-kettunen-nc: Unnecessary question, singles and doubles are both used in the codebase

src/auth/useAuth.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  const logout = useCallback(() => {
+    dispatch(clearApiToken());
+    dispatch(clearUser());
+    setLoggedIn(false);
+    oidcLogout();
+  }, [oidcLogout, dispatch]);
+
+  const isRenewing = useCallback(() => {
+    oidcIsRenewing
```
juho-kettunen-nc, line 72: Here we have the fix incoming, so that's ok

juho-kettunen-nc (approved):

src/auth/useAuth.ts
NC-jsAhonen: Normally _error is null, so if we want to log possible errors from getStoreApiTokens, it could be logged inside an `if (_error) {}` block. Might help debugging, but not necessary.

NC-jsAhonen (approved):

PR #525, opened 2024-09-25
refactor sso provider selection

henrinie-nc:
avoid using boolean values in envvars as yaml, ansible, .env evaluate them differently

### DIFF
--- a/src/auth/constants.ts
+++ b/src/auth/constants.ts
@@ -1,5 +1,7 @@
 import type { LoginProviderProps } from 'hds-react';
 
+type OidcProviderName = 'tunnistamo' | 'tunnistus';
+
 // Tunnistamo SSO (legacy)
 const loginProviderTunnistamoProperties: LoginProviderProps = {
   userManagerSettings: {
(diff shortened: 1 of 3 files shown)


src/auth/constants.ts
```diff
@@ -31,6 +33,9 @@ const loginProviderTunnistusProperties: LoginProviderProps = {
   sessionPollerSettings: { pollIntervalInMs: 300000 } // 300000ms = 5min
 };
 
-export const useTunnistamoOpenIdConnect = import.meta.env.VITE_USE_TUNNISTAMO_OPENID_CONNECT === 'true' || import.meta.env.VITE_USE_TUNNISTAMO_OPENID_CONNECT === true;
+export const oidcProviderName: OidcProviderName = import.meta.env.VITE_OIDC_PROVIDER || 'tunnistus';
```
NC-jsAhonen, line 36: 'tunnistamo' string is assigned to an env var, but 'tunnistus' is just a plain string without a reference to a variable. Looks a bit odd, but is there a reason for it? Could it be a bit cleaner if 'tunnistus' could also be defined in variable?

NC-jsAhonen (changes requested):
The code changes look ok to me. However, I recommend adding an explanation to Tunnistus vs Tunnistamo to the README. Those words are so similar that they are easily confused, and even when not, it is not clear what the difference is and what they are referring to. Will this documentation be added now or in a later PR?

juho-kettunen-nc (approved):

src/auth/constants.ts
henrinie-nc: It is defined in VITE_OIDC_PROVIDER, `|| 'tunnistus'` is a fallback default value.
How do you suggest to change it?

henrinie-nc:
> The code changes look ok to me. However, I recommend adding an explanation to Tunnistus vs Tunnistamo to the README. Those words are so similar that they are easily confused, and even when not, it is not clear what the difference is and what they are referring to. Will this documentation be added now or in a later PR?

I don't intend to document it. My reasoning is that it should be pretty clear from the `constants.ts` file itself what they mean, both options point to urls of those services.

src/auth/constants.ts
NC-jsAhonen: It could be defined as `DEFAULT_PROVIDER_NAME`. But I'm not sure, doesn't feel necessary, because when I think about it more, the string 'tunnistus' is used only here. Maybe defining it in a variable becomes necessary only if the exact 'tunnistus' string is used in several places, to avoid the pitfall of mistyping. So maybe there is no need to define it in a variable in these changes.

NC-jsAhonen (approved):

henrinie-nc:
There is a type that has both options, it should check for mistyping. Ofc it wont do that in envvars or dotenv files.

PR #532, opened 2024-10-10
Receivable types: increase response limit to 50

NC-jsAhonen:
The default limit for `receivable_type/` request is 30, but there are more receivable types for AKV than 30. The limit of response objects is increased to 50, so that all AKV receivable types could be shown.


juho-kettunen-nc:
How many receivable types are there currently for AKV? Could we raise the limit to even higher, so that this issue doesn't happen again?

juho-kettunen-nc (approved):

PR #533, opened 2024-10-10
Mvj 536 map tool icons

henrinie-nc:
What this does to fix certain images for leaflet-draw, is that it copies the images of leaflet-draw to the build.
Since _in current version of leaflet-draw_ its css points the images to `image/xxx` but vite wants them to be in `assets`, we will transform the css `url()`'s that start with `images/` to `assets/images/`.

If you know of a better way let me know. Would maybe prefer inline of the images base64 in the css or something, but not sure how to accomplish it with this specific library. Maybe it is not worth any extra time to spend more fighting it.

On top of this, I also changed some css `url()`'s to point to `/assets/` instead of `../assets` - seems more correct and didn't see any issues with that. Those svg's seems to endup in the css files as inline in any case.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "react-dual-listbox": "^2.0.0",
     "react-foundation": "^0.9.7",
     "react-leaflet": "^2.8.0",
-    "react-leaflet-draw": "^0.19.0",
+    "react-leaflet-draw": "0.19.0",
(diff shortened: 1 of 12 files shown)


juho-kettunen-nc (approved):
No improvement suggestions for this. Good that you also updated the relative paths to absolute ones.

PR #534, opened 2024-10-11
receivable types: fetch beyond page limit

NC-jsAhonen:
The default limit for receivable_type/ request is 30, but there are more receivable types for AKV than 30. This change makes the request in the mvj-ui to make several requests in order to get all the available receivable types for the given service unit.

### DIFF
--- a/src/leaseCreateCharge/requests.ts
+++ b/src/leaseCreateCharge/requests.ts
@@ -7,10 +7,15 @@ export const fetchAttributes = (): Generator<any, any, any> => {
     method: 'OPTIONS'
   }));
 };
-export const fetchReceivableTypes = (): Generator<any, any, any> => {
+export const fetchReceivableTypes = (nextUrl: string): Generator<any, any, any> => {
   const state = store.getState();
   const lease = getCurrentLease(state);
   const serviceUnit = lease.service_unit;
+  if (nextUrl) {
+    return callApi(new Request(nextUrl, {
+      method: 'GET'
+    }));
+  }
   return callApi(new Request(createUrl(`receivable_type/`, {
     service_unit: serviceUnit.id
   }), {
--- a/src/leaseCreateCharge/requests.ts
+++ b/src/leaseCreateCharge/requests.ts
@@ -1,4 +1,5 @@
 import callApi from "@/api/callApi";
+import callApiPaginated from "@/api/callApiPaginated";
 import createUrl from "@/api/createUrl";
 import { store } from "@/index";
 import { getCurrentLease } from '@/leases/selectors';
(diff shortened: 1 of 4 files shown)


NC-jsAhonen:
Sorting the receivable types by name in the saga does not work. I think the solution is to sort the choices in the select component.

juho-kettunen-nc:
Should this PR be reviewed yet because it is in the draft phase, or after the sorting is added?

juho-kettunen-nc:
Great. In my opinion this should be merged in addition to #749 in the API repo, to fix the issue now, and be future-proof if more than 50 are added.

juho-kettunen-nc:
@henrinie-nc  what do you think?

src/leaseCreateCharge/requests.ts
```diff
@@ -7,10 +7,15 @@ export const fetchAttributes = (): Generator<any, any, any> => {
     method: 'OPTIONS'
   }));
 };
-export const fetchReceivableTypes = (): Generator<any, any, any> => {
+export const fetchReceivableTypes = (nextUrl: string): Generator<any, any, any> => {
   const state = store.getState();
   const lease = getCurrentLease(state);
   const serviceUnit = lease.service_unit;
+  if (nextUrl) {
```
henrinie-nc, line 14: Is it possible to implement the complete logic in this function for paginating all the results?

src/leaseCreateCharge/saga.ts
```diff
@@ -30,28 +30,35 @@ function* fetchAttributesSaga(): Generator<any, any, any> {
...
+  let nextUrl = ''
+  const allReceivableTypes = []
   try {
-    const {
-      response: {
-        status: statusCode
-      },
-      bodyAsJson
-    } = yield call(fetchReceivableTypes);
+    while (typeof nextUrl === 'string') {
```
henrinie-nc, line 36: Is this enough of a check to see whether to continue? Maybe something else could be checked to ensure it doesn't end up in infinite loop

juho-kettunen-nc: If there are more pages, `bodyAsJson.next` will be a string that contains the URL. If there are no pages left, `bodyAsJson.next` will be null.

In all cases of the switch case below, the nextUrl will be either a string (should continue), or null (should stop). If there is an error in the `call(fetchRecceivableTypes, nextUrl)` line, the error is caught and nextUrl is again set to null, just in case even though we are out of the loop by then.

The condition could also be `(nextUrl !== null)`, but it should achieve the same thing.

src/leaseCreateCharge/requests.ts
juho-kettunen-nc: Probably. Requires some restructuring of function calls and return values. We would need to restructure the `call` and `callApi` invocations and return values in this function and the saga, because we can't otherwise destructure the response properly here. Using these higher order functions as responses can be a little confusing.

juho-kettunen-nc: It was implemented in the saga because that's the place where we already destructure the JSON data and can easily pick out the next URL.

src/leaseCreateCharge/saga.ts
```diff
@@ -44,7 +44,7 @@ function* fetchReceivableTypesSaga(): Generator<any, any, any> {
       switch (statusCode) {
         case 200:
           allReceivableTypes.push(...bodyAsJson.results);
-          nextUrl = bodyAsJson.next;
+          nextUrl = bodyAsJson.next || null;
```
juho-kettunen-nc, line 47: Is there some value in `bodyAsJson.next` that is not the next URL, or already null if there are no more receivable types?

juho-kettunen-nc: I just thought the value is `null` already, if there's no more data coming from backend.

henrinie-nc: If what Juho says is correct then the or is a bit rendundant right?

src/leaseCreateCharge/requests.ts
henrinie-nc: That does make sense. How about making a version of `callApi` that paginates all the pages? So it could be reused in place of callApi in any of these requests

juho-kettunen-nc (approved):
Looks good

PR #537, opened 2024-10-29
MVJ-312 - Prioritize unarchived areas and primary addresses in the LeaseInfo header element

juho-kettunen-nc:
Before, UI showed the first results of each that were unsorted by
backend. This meant that LeaseInfo could show area identifier and
address from an archived area, while unarchived areas would exist.

TODO: think with customer if feature should be implemented in backend instead, to align invoicing with UI.


henrinie-nc (approved):
Looks good, thanks for improving typing.
You can decide whether it should be implemented in UI or backend, although a more universal solution is better, e.g. if backend would solve this better.

juho-kettunen-nc:
Rebased over develop, to refresh the branch with last 3 weeks' changes.

juho-kettunen-nc:
Again rebased over develop

juho-kettunen-nc:
This PR is no longer about ticket MVJ-312, because the ordering or LeaseArea and LeaseAreaAddress will be performed on the API side: https://github.com/City-of-Helsinki/mvj/pull/790

PR #541, opened 2024-11-07
intendeduse select to use is_active filter by default

henrinie-nc:


juho-kettunen-nc:
There was some discussion in daily about this topic: in some parts of the application we want to see also the inactive intended uses: 
- advanced search
- reports
- django-admin

The only place where the inactives need to be filtered out is the summary edit page.

Is this default problematic then?

henrinie-nc:
> There was some discussion in daily about this topic: in some parts of the application we want to see also the inactive intended uses:
> 
>     * advanced search
> 
>     * reports
> 
>     * django-admin
> 
> 
> The only place where the inactives need to be filtered out is the summary edit page.
> 
> Is this default problematic then?

This function is used by the select component for intendeduse, and that is used in places where it is filtered by service unit, e.g. on the lease page.
It should not affect the advanced search, which gets all the intended uses from the OPTIONS.

juho-kettunen-nc (approved):

PR #543, opened 2024-11-11
Sort radionbuttoninline field choices

NC-jsAhonen:
If the radiobuttoninline field is 'hakija', the choices will be sorted by value so that Yritys is first and Henkilö is last.

### DIFF
--- a/src/application/helpers.ts
+++ b/src/application/helpers.ts
@@ -398,7 +398,13 @@ export const prepareApplicationForSubmission = (sections: Record<string, any>): 
   }
 };
 
-export const getRadioButtonInlineFieldChoicesSorted = (field: FormField, choices: FormFieldChoice[]): FormFieldChoice[] => {
+/**
+ * Sorts the choices of a field based on the criteria specific to different field identifiers.
+ * @param {FormField} field
+ * @param {FormFieldChoice} choices
+ * @returns {FormFieldChoice[]}
+ */
+export const getFieldChoicesSorted = (field: FormField, choices: FormFieldChoice[]): FormFieldChoice[] => {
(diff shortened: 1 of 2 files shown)


src/util/helpers.tsx
```diff
@@ -426,6 +426,10 @@ export const sortByOptionsDesc = (a: Record<string, any>, b: Record<string, any>
   return sortStringDesc(valA, valB);
 };
 
+export const sortBySortOrderAsc = (a: Record<string, any>, b: Record<string, any>): number => {
+  return sortNumberByKeyAsc(a, b, 'sort_order');
```
henrinie-nc, line 430: What is going to happen when `sort_order` is not set, as it can't be set from the form editor (like mentioned in the API PR). The default value seems to be 0, so if there was any ordering before, isn't this going to invalidate it and make it random?

src/application/helpers.ts
```diff
@@ -396,4 +396,11 @@ export const prepareApplicationForSubmission = (sections: Record<string, any>):
     console.log(e);
     return null;
   }
-};
\ No newline at end of file
+};
+
+export const getRadioButtonInlineFieldChoicesSorted = (field: FormField, choices: FormFieldChoice[]): FormFieldChoice[] => {
```
juho-kettunen-nc, line 401: I would add a documentation comment here, saying what is the goal of the function in this case. Or maybe even better to rename the function to `getRadioButtonApplicantFieldChoicesSorted`, to make it clear from the outside that it is only intended for the applicant field and no other use cases.

juho-kettunen-nc (changes requested):
See my comment about function name/documentation.

juho-kettunen-nc:
Good idea to name the function with intention to make it expandable.

juho-kettunen-nc (approved):

PR #544, opened 2024-11-14
MVJ-550 override_receivable_type in ContractRent (UI)

juho-kettunen-nc:
This implements the UI part of override receivable type, where it is a property of **ContractRent**, instead of Rent as the original implementation.


juho-kettunen-nc:
Customer approved the Rent implementation instead of this. Closing.

PR #546, opened 2024-11-22
MVJ-569 - Remove seasonal dates feature from rents tab

juho-kettunen-nc:
This feature is not used at all currently or future, and historically only in one (1) lease. Users want to get rid of it.


NC-jsAhonen (approved):
Approved

PR #547, opened 2024-11-22
Develop to master (UI)

juho-kettunen-nc:
I will rebase locally and not use Github buttons, to avoid duplicate commits in future.


juho-kettunen-nc:
Was no need for rebase, just the regular fast-forward merge

PR #550, opened 2024-12-09
Prettier formatting

NC-jsAhonen:
In some places, the code formatting is a bit off, and some rows are very difficult to read because they are so long.

These changes enable format-on-save feature on mvj-ui repo according to prettier defaults. Please try and see if it works for you, and suggest any changes that come to mind.

It might be necessary to format the whole repo, but I haven't done it yet for this PR yet.

Please test ignoring Robert's "initial to typescript conversion" by running
```bash
git config blame.ignoreRevsFile .git-blame-ignore-revs
```
to set the ignore file in your local git settings, and restarting VSCode. I couldn't make it to work, but perhaps you can.


juho-kettunen-nc:
Awesome, this will make working in this repository much smoother. Will try later.

juho-kettunen-nc:
What settings would you prefer in the `.prettierrc` for JS files? Maybe the same settings as we currently have in `.editorconfig`? 
- indent with spaces
- unix line endings
- indent size 2 spaces
- trim trailing whitespace
- insert final newline

Additionally, looks like the Prettier extension does the following by default, so it would be good to write them out in the config file so we don't get surprised later:
- add dangling comma
- semicolons `;` at the end of each statement
- double quotes `"` for strings instead of single quotes `'`

Which line length do you prefer? Looks like prettier enforces 80 by default. This is of course a safe default, but we could also allow for 100 or 120 if we prefer.

Should we continue using the editorconfig file, or remove it and only use prettierrc? Prettier is expected to understand the editorconfig file, and should override them with those in prettierrc in case of double definitions.

NC-jsAhonen:
@juho-kettunen-nc I added explicit configs

juho-kettunen-nc:
Thanks, I like these settings. What about you @henrinie-nc  ?

juho-kettunen-nc:
And @NC-jsAhonen , do you have additional settings in mind, or how do you like these proposed settings?

juho-kettunen-nc:
I tested the new settings on the file `src/leases/helpers.ts`. The line length setting will influence number of lines in the file, due to new line wrapping:
- no formatter: ~2900 lines
- 80 chars per line: ~3900 lines
- 100 chars per line: ~3600 lines

I liked 100 chars per line, as in the latest commit.

juho-kettunen-nc (approved):

henrinie-nc:
Lets configure prettier to format only ts and tsx files and reformat the project.

.vscode/settings.json
```diff
@@ -0,0 +1,8 @@
+{
```
henrinie-nc, line 1: Maybe add this as an example file like https://github.com/City-of-Helsinki/mvj-ui/blob/develop/.vscode/launch.example.json

juho-kettunen-nc:
> Please test ignoring Robert's "initial to typescript conversion" by running
> 
> ```shell
> git config blame.ignoreRevsFile .git-blame-ignore-revs
> ```
> 
> to set the ignore file in your local git settings, and restarting VSCode. I couldn't make it to work, but perhaps you can.

This works for me, it ignores both your prettier formatting and Robert's typescript conversion, but only when the ignored commit is NOT the first commit of that line's timeline.

For example, line 9 of `src/api/callApi.ts` was changed in both your and Robert's commit, but I only see blame from 6 years ago. On the other hand, some other file whose history starts with the typescript conversion will show blame for the initial commit, even if that commit is on the ignore list.

henrinie-nc (approved):
good to go

.vscode/settings.json
NC-jsAhonen: Done

PR #554, opened 2024-12-11
rent tab: edit old_dwellings_in_housing_companies_price_index_type

NC-jsAhonen:
Tasotarkistus type is now editable


juho-kettunen-nc:
Good progress! I however want to address the naming: the type field is not for index type. I don't think there exists multiple types for the index that would be available for choosing.

I see that the correct interpretation is already in place in docstring of the enum's "In Finnish"-section, which says that the enum is for "Tasotarkistuksen tyyppi". The name of the property would probably then be `periodic_rent_adjustment_type`, and enum `PeriodicRentAdjustmentType`.

juho-kettunen-nc:
But I know there is time pressure to show this to customer tomorrow, so I'll create a ticket about changing the name later this week. This way we can get this merged and displayed tomorrow, as this does not affect what is shown in UI.

juho-kettunen-nc (approved):

PR #560, opened 2024-12-19
Add launch configuration for VS Code internal debugger

juho-kettunen-nc:
Allows you to debug inside VS Code, if you prefer that over browser tools.


henrinie-nc:
Should we add tooling specific configs in the repo? What if everybody want to have their own, it is a bit of a fight against git with the changes

juho-kettunen-nc:
Renamed file to be an example, to avoid conflicts with each developer's personal launch.json configurations.

henrinie-nc (approved):

PR #562, opened 2024-12-20
tasotarkistus: add ui data to start date and review days

NC-jsAhonen:
Add two more info balls to the tasotarkistus section: start date and review days.


src/leases/enums.ts
```diff
@@ -933,6 +933,9 @@ export const LeaseRentDueDatesFieldTitles = {
 export const LeaseRentOldDwellingsInHousingCompaniesPriceIndexFieldPaths = {
   OLD_DWELLINGS_IN_HOUSING_COMPANIES_PRICE_INDEX_TYPE: 'rents.child.children.old_dwellings_in_housing_companies_price_index_type',
   POINT_FIGURES: 'rents.child.children.old_dwellings_in_housing_companies_price_index.point_figures',
+  // The following two are for the UI data purposes only
```
henrinie-nc, line 936: Is this a common pattern not to name these UiData ones differently?

NC-jsAhonen: For START_DATE, the data is derived from lease.start_date directly, but if I mapped the UI data to that field, the info ball would show the same text as in other places the lease.start_date is used. But I decided to map it to a different path to make it possible to create separate notes for it in the tasotarkistus view

For REVIEW_DAYS, it is not a proper form field with data of its own, but it is derived from the lease start date.

henrinie-nc (approved):

PR #563, opened 2024-12-23
rename tasotarkistus type variables

NC-jsAhonen:
Previously, tasotarkistus type for rent was named as `old_dwellings_in_housing_companies_price_index_type`, but in this change, it is renamed as `periodic_rent_adjustment_type`.
 
The related API changes are here: https://github.com/City-of-Helsinki/mvj/pull/801


henrinie-nc:
With this change, is there any potential misunderstandings that could arise from other index adjustments?

juho-kettunen-nc (approved):

juho-kettunen-nc:
> With this change, is there any potential misunderstandings that could arise from other index adjustments?

@henrinie-nc  Do you mean the model IndexAdjustedRent? First time I see that model. No in my opinion, because
- PeriodicRentAdjustment and IndexAdjustedRent are in separate sections in the UI, and are separate models
- IndexAdjustedRent does apparently doesn't have a relation to an index, only the details for the adjustment amount.
- This PR fixes the implication that the used index has different "types", when in reality the "type" is for periodic rent adjustment as a whole and not for the index it uses under the hood.

PR #565, opened 2024-12-27
form editor: allow overflow in modal

NC-jsAhonen:


henrinie-nc:
Did you find any reason for why did this break?
I wonder if it would be useful to look into that a bit more to see if there is some issue that is affecting this

henrinie-nc (approved):
Lets go with this

PR #567, opened 2025-01-09
add duplicate commit to git blame ignore revs

NC-jsAhonen:


juho-kettunen-nc:
I don't see the original ignored commit `d01bd7f80124db5f9455222ea2a2599590e37465` anywhere anymore. Can the newer formatting commit replace that line, or?

henrinie-nc (approved):

NC-jsAhonen:
I could find both of them via `git log`, so I will be including both of them.

PR #571, opened 2025-02-04
Ignore launch.json to allow personal configurations

juho-kettunen-nc:


juho-kettunen-nc:
Hmm, need to also ignore `settings.json`

PR #572, opened 2025-02-05
MVJ-131 ClamAV (UI)

juho-kettunen-nc:
Pass filescan error messages to user. Added a long timeout so that user has time to read it.

Might require more changes but not sure yet.


juho-kettunen-nc:
Error message translation doesn't work in the backend yet

PR #575, opened 2025-02-24
replace sonarcloud action with sonarqube drop-in replacement

henrinie-nc:
"Warning: This action is deprecated and will be removed in a future release. Please use the sonarqube-scan-action action instead. The sonarqube-scan-action is a drop-in replacement for this action."


NC-jsAhonen (approved):
Approved

PR #577, opened 2025-02-27
Mvj 642 reports link to lease new tab and styling

NC-jsAhonen:
Statistic report, link to lease column changes:
- open in new tab
- unset margin and border to match the spacing of other cells


juho-kettunen-nc (approved):
Works as intended

PR #581, opened 2025-03-10
MVJ-203 Unselect areasearch preparer when lessor is changed

juho-kettunen-nc:
The requested change was implemented, but includes a bug: preparer is unset even if nothing is changed in the form.

This bug was present in the list view editing modal already, but now added to the areasearch details edit form. TODO

### DIFF
--- a/src/areaSearch/components/AreaSearchApplicationEdit.tsx
+++ b/src/areaSearch/components/AreaSearchApplicationEdit.tsx
@@ -5,7 +5,7 @@ import flowRight from "lodash/flowRight";
 import orderBy from "lodash/orderBy";
 import get from "lodash/get";
 import { Column, Row } from "react-foundation";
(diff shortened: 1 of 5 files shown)


juho-kettunen-nc:
Fixed bug for areasearch details page: preparer value is now preserved if the field is not touched. The bug still remains for areasearch list page editing modal.

henrinie-nc (approved):
Looks good to me, approving draft to indicate this part has been reviewed, if further fixes come remind me to review those

juho-kettunen-nc:
@henrinie-nc  fixed the list page editing modal, and changed preparer identification from `id` to `value`, because value corresponds to id in the select component, and value is always present, whereas id is not present if selection is not changed.

henrinie-nc (approved):
Good to go

PR #584, opened 2025-03-19
MVJ-610 Tasotarkistus index point figure update

juho-kettunen-nc:
Add new field to Rent: start_price_index_point_figure_year.

Corresponding API PR: https://github.com/City-of-Helsinki/mvj/pull/871

### DIFF
--- a/src/leases/helpers.ts
+++ b/src/leases/helpers.ts
@@ -65,6 +65,7 @@ import type { CommentList } from "@/comments/types";
 import type { Attributes, LeafletFeature, LeafletGeoJson } from "types";
 import type { RootState } from "@/root/types";
 import type { LeaseList, DueDate } from "@/leases/types";
+import type { IndexPointFigureYearly } from "@/oldDwellingsInHousingCompaniesPriceIndex/types";
 
 /**
  * Test is lease empty
@@ -2343,6 +2344,10 @@ export const getContentRents = (lease: Lease): Array<Record<string, any>> =>
         old_dwellings_in_housing_companies_price_index:
           rent.old_dwellings_in_housing_companies_price_index,
         periodic_rent_adjustment_type: rent.periodic_rent_adjustment_type,
+        start_price_index_point_figure_value:
+          rent.start_price_index_point_figure_value,
+        start_price_index_point_figure_year:
+          rent.start_price_index_point_figure_year,
       };
     })
     .sort(sortByStartAndEndDateDesc);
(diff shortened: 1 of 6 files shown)


src/leases/helpers.ts
```diff
@@ -3968,3 +3973,45 @@ export const isATypedLease = (leaseTypeIdentifier: string): boolean => {
...
+ * lease's start date, e.g. 2023 if lease starts in 2024.
+ */
+const getPointFigureYearAndValueFromLease = (
+  pointFigures: IndexPointFigureYearly[],
+  leaseStartDate: string,
+): { value?: number; year?: number } => {
+  const leaseStartYear = new Date(leaseStartDate).getFullYear();
+  const figure =
+    pointFigures?.find(
+      (x: IndexPointFigureYearly) => x.year == leaseStartYear - 1,
```
NC-jsAhonen, line 4013: `x` could be named as `point_figure`

src/leases/helpers.ts
```diff
@@ -3968,3 +3973,45 @@ export const isATypedLease = (leaseTypeIdentifier: string): boolean => {
...
+): string => {
+  if (yearFromRent && valueFromRent)
+    return `${yearFromRent} * ${valueFromRent}`;
+
+  const { year: yearFromLease, value: valueFromLease } =
+    getPointFigureYearAndValueFromLease(pointFigures, leaseStartDate);
+  if (yearFromLease && valueFromLease)
+    return `${yearFromLease} * ${valueFromLease}`;
+
+  return "Indeksipisteluku puuttuu";
```
NC-jsAhonen, line 3996: Should there be a translation for this, instead of a raw string in Finnish?

NC-jsAhonen (changes requested):
Looks basically good. Could you make changes or reply to my comments. They are about small tweaks that could improve the code a little bit.

src/leases/helpers.ts
juho-kettunen-nc: Agreed

src/leases/helpers.ts
juho-kettunen-nc: Yes, I'll use the i18 for this

juho-kettunen-nc: Even if swedish is not much used here, I'll create the translation keys for future.

juho-kettunen-nc: Ahh, I forgot that we don't even use translations in officer UI! Every text that is specified in frontend is hardcoded to finnish, for example in src>leases>enums.ts @NC-jsAhonen

src/leases/helpers.ts
juho-kettunen-nc: Fixed in 34fcee771926e4ff2abb074e437307d907d10bc2

src/leases/helpers.ts
NC-jsAhonen: Ok, let's go with the raw string then.

PR #593, opened 2025-04-09
increase limit for FieldTypeIndendedUseSelect

henrinie-nc:
in order to fetch all intended uses


juho-kettunen-nc (approved):
Gotta fetch them all

PR #594, opened 2025-04-09
use service_unit in contact_exists check

henrinie-nc:


src/contacts/types.ts
```diff
@@ -17,7 +18,26 @@ export type ContactState = {
 export type ContactExistsResponse = {
   exists: boolean;
 };
-export type Contact = Record<string, any>;
+export type Contact = {
+  service_unit: number;
+  type: (typeof ContactTypes)[keyof typeof ContactTypes];
```
juho-kettunen-nc, line 23: How does this work? Never seen such syntax

juho-kettunen-nc: I guess it just limits the possible values to enum values? I wish there was a simpler way to express this. If not, "just TS things".

src/landUseContract/components/sections/LitigantsEdit.tsx
```diff
@@ -246,7 +246,9 @@ class TenantsEdit extends PureComponent<Props, State> {
               : null;
 
             if (contactIdentifier && !isEmptyValue(contactIdentifier)) {
-              const exists = await contactExists(contactIdentifier);
+              const exists = await contactExists({
+                identifier: contactIdentifier,
+              });
```
juho-kettunen-nc, line 251: Is there no need to stay within service unit contacts in Litigants and Tenants at this step of the code? Or just out of scope of this ticket?

juho-kettunen-nc (approved):
Approved. Have some questions I'd like clarified.

src/contacts/types.ts
henrinie-nc: typeof gets the type of the ContactTypes without having to create one (as it can be determined). So that could be replaced with `ContactTypesType` and then it would become `ContactTypesType[keyof ContactTypesType]` https://www.typescriptlang.org/docs/handbook/2/indexed-access-types.html

And as I understand it, then it is created as sort of a literal with the values in this object.

src/landUseContract/components/sections/LitigantsEdit.tsx
henrinie-nc: Great question that I didn't even think of as it was not directly in the scope, but if making such changes I need to also consider this. Thanks for pointing it out

src/contacts/types.ts
henrinie-nc: Oh forgot to mention about keyof, it makes a union of the keys, so basically this selects all the values of the keys

src/leases/enums.ts
```diff
@@ -120,7 +121,7 @@ export const Classification = {
   PUBLIC: "public",
   CONFIDENTIAL: "confidential",
   OFFICIAL: "official",
-};
```
juho-kettunen-nc, line 123: Good idea to enforce immutability further, I did not remember existence of `as const`.

juho-kettunen-nc (approved):

src/leases/enums.ts
henrinie-nc: It is just required for getting the typing to literal working, I assume so that typescript can know it won't change

PR #596, opened 2025-04-16
CreateLeaseModal in AreaSearchApplicationEdit

NC-jsAhonen:
A modal for creating a lease in an area search application view.

No data from the areasearch is yet transferred into the new lease. Those are implemented in the upcoming PRs.

I also created an enum for the create lease button. I followed the wording "Luo vuokraus" for the button in the area search application view as is specified on the ticket. I kept the "Luo vuokraustunnus" wording for other locations.


juho-kettunen-nc (approved):
Aligns with subticket requirements

PR #602, opened 2025-04-28
fix serviceunit typing for Contact

henrinie-nc:


juho-kettunen-nc (approved):

henrinie-nc:
I am going to close this, it is not fixing the root cause. The value seems to change between Contact and ContactForm at some point from ServiceUnit to id.

PR #603, opened 2025-04-29
MVJ-836 Areasearch map tab without geometry

juho-kettunen-nc:
Suggestions welcome: Is there need for a prettier presentation than plain text as a placeholder in the map tab when geometry is missing? Picture in ticket.

### DIFF
--- a/src/areaSearch/components/map/SingleAreaSearchMap.tsx
+++ b/src/areaSearch/components/map/SingleAreaSearchMap.tsx
@@ -61,7 +61,7 @@ class SingleAreaSearchMap extends Component<Props> {
     const { geometry, minimap = false } = this.props;
 
     if (!geometry) {
-      return <>{"Ei geometriaa."}</>;
+      return <p>{"Ei geometriaa."}</p>;
     }
 
     const center = this.getMapCenter();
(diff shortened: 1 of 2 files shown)


src/util/map.ts
```diff
@@ -204,6 +204,10 @@ export const getCenterFromCoordinates = (
 export const getAreaFromGeoJSON = (geometry: Record<string, any>): number => {
   let area = 0;
 
+  if (!geometry || !geometry.type) {
```
henrinie-nc, line 207: I guess `!geometry?.type` would suffice too

src/areaSearch/components/map/SingleAreaSearchMap.tsx
```diff
@@ -59,6 +59,11 @@ class SingleAreaSearchMap extends Component<Props> {
 
   render(): JSX.Element {
     const { geometry, minimap = false } = this.props;
+
+    if (!geometry) {
+      return <>{"Ei geometriaa."}</>;
```
henrinie-nc, line 64: Maybe some element could be surrounding the text, e.g. `<p>` or `<div>`. What do you think?

henrinie-nc (approved):

src/areaSearch/components/map/SingleAreaSearchMap.tsx
juho-kettunen-nc: Yeah, I didn't see much of a difference, but that would probably be semantically better.

src/util/map.ts
juho-kettunen-nc: Right, forgot this is JavaScript.

juho-kettunen-nc:
Ehh, I tried to rebase-merge to develop on command line and to simultaneously combine the 4 commits into two. Develop branch was updated just fine, but now this PR's commit hashes are out of date. I'll close this.

PR #609, opened 2025-05-23
create contact in area search application: fix country field default value

NC-jsAhonen:
Fix the issue of having a wrong kind of string in the country field. The country field in the application form is a text field, and has the string "Suomi" there by default in the mvj-ui-public. However, in the leasing_contact, it is a field that expects country codes. 

This workaround resolves the string "suomi" (case insensitive) as "FI", so the default value is Finland in the create contact form in most cases. If the string is something else, the default value becomes null, and needs to be set manually in the create contact form, or the contact will have the country field as empty when it is created.

There is a bug ticket for fixing the country and language fields for the application forms so that the string mappings match those of contacts.

### DIFF
--- a/src/areaSearch/helpers.ts
+++ b/src/areaSearch/helpers.ts
@@ -262,12 +262,16 @@ export const getContactFromAnswerFields = (
       EMPTY_DEFAULT_FIELD,
     )?.value;
 
-    const { address_protection, language, postal_code } =
+    const { address_protection, country, language, postal_code } =
       MapFormAnswerFieldsToContactFields;
 
     switch (MapFormAnswerFieldsToContactFields[key]) {
       case address_protection: {
         contact.address_protection = getAddressProtection(value);
+        break;
+      }
+      case country: {
+        contact.country = value?.toLowerCase() === "suomi" ? "FI" : null;
         break;
       }
       case language: {


src/areaSearch/helpers.ts
```diff
@@ -262,14 +262,18 @@ export const getContactFromAnswerFields = (
...
+    const { address_protection, country, language, postal_code } =
       MapFormAnswerFieldsToContactFields;
 
     switch (MapFormAnswerFieldsToContactFields[key]) {
       case address_protection: {
         contact.address_protection = getAddressProtection(value);
         break;
       }
+      case country: {
+        contact.country = value?.toLowerCase() === "suomi" ? "FI" : null;
```
juho-kettunen-nc, line 274: I would, just in case, compare against all `suomi`, `finland` `finnland` to catch the most usual spellings of our country. This would accommodate cases where a multilingual user wants to replace the default value, even if it means the same thing in the end. Like:

```
const spellings = ["suomi", "finland", "finnland"];
contact.country = spellings.includes(value?.toLowerCase()) ? "FI" : null;
```

juho-kettunen-nc (approved):
This works. You can consider my suggestion if you think it's reasonable.

PR #621, opened 2025-07-31
MVJ-886: credit decision to final-form

NC-jsAhonen:
Here are refactorings of credit decision views from redux-form to final-form.

There is also a refactor from the old class syntax to the newer hook syntax, to make future changes easier.


juho-kettunen-nc (approved):
Looks good

PR #622, opened 2025-07-31
MVJ-881 SAP invoices to final form and hooks

NC-jsAhonen:


src/sapInvoice/components/SapInvoicesListPage.tsx
```diff
@@ -133,176 +131,106 @@ const getColumns = (invoiceAttributes: Attributes) => {
...
-      if (!this._hasFetchedInvoices) {
-        // No search has been done yet
+      if (!hasFetchedInvoices) {
         handleSearch();
-        this._hasFetchedInvoices = true;
-      } else if (
-        userActiveServiceUnit !== prevUserActiveServiceUnit &&
-        !currentSearch.includes("service_unit")
-      ) {
-        // Search again after changing user active service unit only if not explicitly setting the service unit filter
```
juho-kettunen-nc, line 234: Was this behavior preserved? In the new code I don't see comparisons between the new selected service unit and the previous.

src/sapInvoice/components/SapInvoicesListPage.tsx
```diff
@@ -133,176 +131,106 @@ const getColumns = (invoiceAttributes: Attributes) => {
...
+        setHasFetchedInvoices(true);
+      } else if (!currentSearch.includes("service_unit")) {
         handleSearch();
       }
     }
 
-    if (currentSearch !== prevSearch) {
```
juho-kettunen-nc, line 239: Is this also implemented by final-form? I think we wanted to avoid unnecessary fetches if nothing changes in the filters.

NC-jsAhonen: Yes, this logic is replaced by the dependency array of the useEffect

NC-jsAhonen: So the dependency array is this: [location.search, userActiveServiceUnit]

PR #625, opened 2025-08-11
MVJ-902 use lease_area draft_geometry field

NC-jsAhonen:
Instead of using a separate `lease_area_draft` object, use the `draft_geometry` of each `lease_area`.

API PR: https://github.com/City-of-Helsinki/mvj/pull/971


src/components/map/MapInput.tsx
```diff
@@ -75,7 +83,7 @@ class MapInput extends Component<Props, State> {
     layer.showMeasurements();
     const features = [];
     this.state.featureGroup?.leafletElement.eachLayer((layer) => {
-      layer.bindPopup("Alueluonnos");
+      layer.bindPopup(renderPopupOrTooltip(layer?.feature?.properties?.identifier));
```
juho-kettunen-nc, line 86: Risk of undefined at runtime?

PR #626, opened 2025-08-14
credit decision search form: fix rendering issue

NC-jsAhonen:
By removing the empty `subscription` object (and thus enabling subscription), the form now re-renders on changes. Secondly, make `values.contact_type` as conditionally chained, we ensure that there will be no error in case that the values object happens to be undefined.


henrinie-nc:
`export default flowRight(SearchForm) as React.ComponentType<any>;` <- flowRight does nothing here and can be removed

PR #630, opened 2025-08-20
Mvj 1013 lease summary history final form

henrinie-nc:
Thought this had some redux-form since it used to initialize a form, but turns out that initialize was not needed at all.
Intention was to migrate to final-form but nothing to migrate here after converting to functional component :/


NC-jsAhonen (approved):
lgtm

PR #631, opened 2025-08-21
fix LeasePage having stale props in saveUnsavedChanges due to interval

henrinie-nc:
This prevented saveUnsavedChanges saving certain forms, because their "dirtiness" didn't update for the timer, as the timer had stale props.
If you find any issues or know better approaches, let me know.

### DIFF
--- a/src/leases/components/LeasePage.tsx
+++ b/src/leases/components/LeasePage.tsx
@@ -1,5 +1,4 @@
 import React, { useEffect, useMemo, useRef, useState } from "react";
-import PropTypes from "prop-types";
 import { connect } from "react-redux";
 import {
   change as reduxFormChange,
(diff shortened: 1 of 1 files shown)


NC-jsAhonen:
I don't have a better solution for handling the stale props.

But is `saveUnsavedChanges` a good term for the function? It "saves" unsaved changes to session storage, which is different from what is done in `saveChanges`, although both are "saving". Should it be `cacheUnsavedChanges` or something similar, to clarify the difference?

NC-jsAhonen (approved):
I don't have a better solution for handling the stale props, so I approved.

But is saveUnsavedChanges a good term for the function? It "saves" unsaved changes to session storage, which is different from what is done in saveChanges, although both are "saving". Should it be cacheUnsavedChanges or something similar, to clarify the difference?

PR #634, opened 2025-09-03
Mvj 1015 lease contacts final form

henrinie-nc:
Transforms Tenants tab in LeasePage to use final-form.
Introduces new packages `final-form-arrays` and `react-final-form-arrays` to replace similar redux-form functionality.


juho-kettunen-nc (approved):
Looks good

PR #678, opened 2026-01-15
Refactor Contract page to final-form

kimi-ihalainen-nc:


src/leases/components/leaseSections/contract/ContractItemEdit.tsx
```diff
@@ -485,6 +485,7 @@ const Collaterals: React.FC<CollateralsProps> = ({
                         key={index}
                         field={field}
                         onRemove={handleRemove}
+                        collateralType={fields.value?.[index]?.type}
```
kimi-ihalainen-nc, line 488: Previously the collateralType inside `<CollateralEdit>` used to come from
```
  const collateralType: number | null | undefined = useSelector((state) =>
    selector(state, `${field}.type`),
  );
```
But the best way I found to replace the selector() was to instead insert it as a prop here in the parent. I feel like this is not the best approach. But I might be wrong about being wrong.
The same change was made in DecisionsEdit on row 77 and InspectionsEdit on row 101.

src/leases/components/leaseSections/contract/CollateralEdit.tsx
```diff
@@ -680,4 +678,5 @@ const CollateralEdit: React.FC<Props> = ({ field, onRemove }) => {
   );
 };
 
+const formName = FormNames.LEASE_CONTRACTS;
```
henrinie-nc: Is this needed anymore?

src/leases/components/leaseSections/contract/ContractsEdit.tsx
```diff
@@ -1,18 +1,18 @@
 import React, { ReactElement, useEffect } from "react";
-import { useDispatch, useSelector } from "react-redux";
-import { FieldArray, reduxForm } from "redux-form";
+import { useSelector } from "react-redux";
+import { FieldArray } from "react-final-form-arrays";
+import { Form } from "react-final-form";
+import { FormApi } from "final-form";
```
henrinie-nc: Seems like this could be `import type { FormApi } from "final-form";`
It mainly would ensure the type will never be included in the build (not needed during runtime)

src/leases/components/leaseSections/contract/DecisionsEdit.tsx
```diff
@@ -1,6 +1,8 @@
 import React, { ReactElement, useEffect, useState } from "react";
-import { connect, useDispatch, useSelector } from "react-redux";
-import { FieldArray, reduxForm } from "redux-form";
+import { useDispatch, useSelector } from "react-redux";
+import { FieldArray } from "react-final-form-arrays";
+import { Form } from "react-final-form";
+import { FormApi } from "final-form";
```
henrinie-nc: Could be type import

src/leases/components/leaseSections/contract/DecisionsMainEdit.tsx
```diff
@@ -18,8 +18,19 @@ import { getUiDataLeaseKey } from "@/uiData/helpers";
 import { isFieldAllowedToRead } from "@/util/helpers";
 import { getAttributes } from "@/leases/selectors";
 import type { Attributes } from "types";
+import { FormApi } from "final-form";
```
henrinie-nc: Could be type import

src/leases/components/leaseSections/contract/InspectionsEdit.tsx
```diff
@@ -1,7 +1,9 @@
-import React, { ReactElement, useEffect } from "react";
-import { useDispatch, useSelector } from "react-redux";
+import React, { ReactElement } from "react";
+import { useSelector } from "react-redux";
 import { Row, Column } from "react-foundation";
-import { FieldArray, reduxForm } from "redux-form";
+import { FieldArray } from "react-final-form-arrays";
+import { Form } from "react-final-form";
+import { FormApi } from "final-form";
```
henrinie-nc: Could be type import

src/leases/components/leaseSections/contract/ContractItemEdit.tsx
henrinie-nc: Seems like this is something we need to do, since this does not end up in redux anymore right?

henrinie-nc (approved):
Everything looks good, small fixes suggested in comments, check those before merging

src/leases/components/leaseSections/contract/CollateralEdit.tsx
kimi-ihalainen-nc: I followed the example of leaving the form names even when they were no longer needed. These can definitely be removed.

henrinie-nc: I was a bit undecided about it, as they could be a link between the LeasePage's logic. But either way is fine to me, to keep or not

PR #689, opened 2026-03-05
Mvj 1181 fix basisofrent mast fields

henrinie-nc:
two commits, one to fix the issue, the other is fixing typing and props passing.
the fix feels like a hack, but it also seems to work. not fully clear how it is supposed to work, it simply didn't get any children for the fieldarray to render.


kimi-ihalainen-nc (approved):
Seems good enough for now. Will likely need to take another look at this when refactoring these files to final form.

PR #690, opened 2026-03-10
Fixes for various invoicing page issues

kimi-ihalainen-nc:
-Added logic for new API errors in invoice sagas
-Made invoicing page to be read only for users in other service units.
-Fixed invoice panel validation state not updating
-Made EditInvoiceForm UI fields match API logic for updating (invoice.py InvoiceViewSet get_serializer_class)
-Fixed invoice panel rows overflowing incorrectly
-Fixed crash caused by having 0 invoice rows

### DIFF
--- a/src/invoices/saga.ts
+++ b/src/invoices/saga.ts
@@ -181,7 +181,6 @@ function* patchInvoiceSaga({
       case 200:
         yield put(fetchInvoicesByLease(bodyAsJson.lease));
         yield put(receivePatchedInvoice(bodyAsJson));
-        yield put(receiveIsEditClicked(false));
         displayUIMessage({
           title: "",
           body: "Lasku tallennettu",
@@ -200,6 +199,17 @@ function* patchInvoiceSaga({
         );
         break;
 
+      case 403:
+        yield put(notFound());
+        yield put(
+          receiveError(
+            new SubmissionError({
+              ...bodyAsJson,
+            }),
+          ),
+        );
+        break;
+
       case 500:
         yield put(notFound());
         yield put(receiveError(new Error(bodyAsJson)));
@@ -209,6 +219,8 @@ function* patchInvoiceSaga({
     console.error('Failed to edit invoice with error "%s"', error);
     yield put(notFound());
     yield put(receiveError(error));
+  } finally {
+    yield put(receiveIsEditClicked(false));
   }
 }
 
--- a/src/leases/components/leaseSections/invoice/Invoices.tsx
+++ b/src/leases/components/leaseSections/invoice/Invoices.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useMemo, useRef, useState } from "react";
+import React, { useEffect, useMemo, useState } from "react";
 import { useDispatch, useSelector } from "react-redux";
 import { ActionTypes, AppConsumer } from "@/app/AppContext";
 import Authorization from "@/components/authorization/Authorization";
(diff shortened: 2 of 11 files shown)


src/leases/components/leaseSections/invoice/Invoices.tsx
```diff
@@ -127,8 +130,13 @@ const Invoices: React.FC = () => {
     ),
   );
   const usersPermissions = useSelector(getUsersPermissions);
+  const activeServiceUnit = useSelector(getUserActiveServiceUnit);
   const dispatch = useDispatch();
 
+  const isServiceUnitSameAsActiveServiceUnit = () => {
+    return activeServiceUnit?.id === currentLease?.service_unit?.id;
+  };
```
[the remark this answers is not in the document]
kimi-ihalainen-nc: I was also considering this, but wasn't sure about implementing it.

src/invoices/saga.ts
```diff
@@ -200,6 +199,17 @@ function* patchInvoiceSaga({
         );
         break;
 
+      case 403:
+        yield put(notFound());
```
henrinie-nc, line 203: what does this do functionally? It is a bit confusing for the reader (I see it is present in case 400 too).

src/leases/components/leaseSections/invoice/forms/EditInvoiceForm.tsx
```diff
@@ -245,6 +255,23 @@ const EditInvoiceForm: React.FC<Props> = ({
 
   const rows = invoice ? invoice.rows : undefined;
 
+  const isSentToSap = invoice?.sent_to_sap_at || invoice?.sap_id;
```
henrinie-nc: Should be converted to boolean, if type is sent to `const isSentToSap: bool = ...` it would likely complain.

src/leases/components/leaseSections/invoice/forms/EditInvoiceForm.tsx
```diff
@@ -245,6 +255,23 @@ const EditInvoiceForm: React.FC<Props> = ({
...
+  const isCreditNote = invoice?.type === InvoiceType.CREDIT_NOTE;
+  const isGenerated = invoice?.generated;
```
henrinie-nc: Probably also should be converted to boolean, unless `generated` is a boolean.

src/leases/saga.ts
```diff
@@ -440,6 +440,17 @@ function* patchLeaseInvoiceNotesSaga({
         );
         break;
 
+      case 403:
```
henrinie-nc, line 443: does this case need the `finally` phase?

src/invoices/saga.ts
kimi-ihalainen-nc: It's a boolean related to error handling in some way on fetching and saving. It is quite confusingly named, but it is used everywhere across sagas on error cases.

src/leases/components/leaseSections/invoice/forms/EditInvoiceForm.tsx
kimi-ihalainen-nc: Will convert all 3 into booleans for consistency.

src/leases/saga.ts
kimi-ihalainen-nc: I am currently implementing these in MVJ-1184 ticket. It's easier to test them on a branch which still allows all fields to be easily edited incorrectly.

kimi-ihalainen-nc: Actually, this doesn't disable the save button on save, which is why it doesn't need to re-enable it.

henrinie-nc (approved):

juho-kettunen-nc (commented):
Looks good

PR #704, opened 2026-03-31
Refactor rent page to functional components

kimi-ihalainen-nc:


kimi-ihalainen-nc:
Opening another one when commits are a bit more settled down.

PR #709, opened 2026-04-08
Landuse prototype: simplify collaterals tab

juho-kettunen-nc:


juho-kettunen-nc:
SonarQube fails with "Failed to restore cache entry". Won't fix in this PR

juho-kettunen-nc:
These changes are pair programmed by me and @henrinie-nc , so I consider them reviewed already.

PR #723, opened 2026-04-24
Implement collateral forms per type

juho-kettunen-nc:
First draft.


juho-kettunen-nc:
TODO
- improve layout of the fields to look better.
- consider the component structure: how much do we want to extract to separate files, and how much to inline in already large files?

henrinie-nc (approved):

PR #730, opened 2026-05-18
Initialize MAST calculator row's area values to 0

juho-kettunen-nc:
We want to keep the backend requirement for non-empty area values. This helps avoid manually filling in the empty rows that are not used in the  calculation at that time.


juho-kettunen-nc:
This PR will not stand as-is, because we found lots of problems with the existing implementation. Closing

PR #737, opened 2026-05-26
add create multiple invoices dialog

henrinie-nc:
- use preset values
- add two invoice rows with default selite


juho-kettunen-nc:
I approve

PR #741, opened 2026-05-28
Yarn v4

henrinie-nc:


Dockerfile
```diff
@@ -4,27 +4,24 @@ FROM registry.access.redhat.com/ubi9/nodejs-24 AS appbase
...
 USER default
+# Yarn version is read from "packageManager" in package.json
+RUN corepack install
+RUN yarn --version
 
 # Install exact versions of dependencies and clean cache.
-# Use non-prod flag to install devDependencies.
-# Ignore engine restrictions for HDS 5.2.0 which requires Node 22, but we use 24.
-RUN yarn --frozen-lockfile --ignore-engines --production=false && yarn cache clean --force
+# Engine restrictions for HDS 5.2.0 are suppressed via ignoredErrors in .yarnrc.yml.
```
juho-kettunen-nc: I don't see `ignoredErrors` entry in .yarnrc.yml, so I expect the install to fail in CI.

.github/workflows/test.yml
```diff
@@ -18,7 +18,8 @@ jobs:
       uses: actions/setup-node@v6
       with:
         node-version: ${{ matrix.node }}
-    - run: yarn install --frozen-lockfile --ignore-scripts
+    - run: corepack enable && corepack install
+    - run: yarn install --immutable
```
juho-kettunen-nc, line 22: We should probably add the script disabling option here, because I don't think .yarnrc.yml is available in this Actions pipeline.

juho-kettunen-nc: I thought wrong, we clone the repo in pipeline, so it should be fine. There's no script disabling option for yarn v4 CLI command, apparently.

Dockerfile
juho-kettunen-nc: There is no such option as `ignoredErrors` in yarn documentation, we found. Fine now because the comment is removed.

juho-kettunen-nc (approved):

PR #765, opened 2026-06-24
refactor trade register components

henrinie-nc:
- make functional
- replace higher order components with hooks
- memo where PureComponent was used to avoid re-rendering with same props


kimi-ihalainen-nc (approved):
Looks clean and good.

PR #766, opened 2026-06-25
Mvj 1300 react foundation

henrinie-nc:
Removes react-foundation
Implements custom grid system that is meant to work very closely to foundation *deprecated* flex-grid https://get.foundation/sites/docs/flex-grid.html
Removes foundation-sites reveal and replaces with HDS Dialog
Removes unused foundation things
Makes more explicit which styling implementations from foundations is being used, so migrating away from it is more clear. **There are commented out components that didnt seem to be used**, if something becomes broken enabling some of them might fix it.


juho-kettunen-nc (approved):
Good stuff

PR #767, opened 2026-06-25
Refactor LeaseListPage to functional components

kimi-ihalainen-nc:


apps/leasing/src/leases/components/LeaseListPage.tsx
```diff
@@ -1,6 +1,11 @@
-import React, { Fragment, PureComponent } from "react";
-import flowRight from "lodash/flowRight";
-import { connect } from "react-redux";
+import React, {
+  useCallback,
```
henrinie-nc: Probably worth checking these, but I think in the other PR targeting this you already fixed these

apps/leasing/src/leases/components/LeaseListPage.tsx
```diff
@@ -120,177 +126,294 @@ const visualizationTypeOptions = [
...
+      if (searchQuery.visualization === VisualizationTypes.MAP) {
+        setVisualizationType(VisualizationTypes.MAP);
+        searchByBBox();
+      } else {
+        search();
       }
-    }
+    };
+
+    if (!userActiveServiceUnit) return;
```
henrinie-nc: Is this needed here at all?

henrinie-nc (approved):
check the imports plus one comment regarding early return in useEffect (is that one needed at all)

apps/leasing/src/leases/components/LeaseListPage.tsx
kimi-ihalainen-nc: I don't know why I didn't spot this; It was needed in one iteration to avoid double fetching. Thanks for noticing it.

kimi-ihalainen-nc:
Fixed mentioned issues

PR #768, opened 2026-06-26
Autoselect user service units in lease list page

kimi-ihalainen-nc:
Converts service unit filter to multiselect and selects user's service units as default values

### DIFF
--- a/apps/leasing/src/leases/components/LeaseListPage.tsx
+++ b/apps/leasing/src/leases/components/LeaseListPage.tsx
@@ -5,7 +5,6 @@ import { Row, Column } from "@/components/grid/Grid";
 import debounce from "lodash/debounce";
 import isArray from "lodash/isArray";
 import isEmpty from "lodash/isEmpty";
-import FieldTypeSelect from "@/components/form/final-form/FieldTypeSelect";
 import SearchInputColumn from "@/components/search/SearchInputColumn";
 import SearchLabel from "@/components/search/SearchLabel";
 import SearchLabelColumn from "@/components/search/SearchLabelColumn";
@@ -82,7 +81,7 @@ import {
   getMethods as getLeaseMethods,
 } from "@/leases/selectors";
 import { getLessorList } from "@/lessor/selectors";
-import { getUserActiveServiceUnit } from "@/usersPermissions/selectors";
+import { getUserServiceUnits } from "@/usersPermissions/selectors";
 import {
   getServiceUnits,
   getIsFetching as getIsFetchingServiceUnits,
@@ -101,6 +100,8 @@ import {
   fetchAttributes as fetchUiDataAttributes,
   fetchUiDataList,
 } from "@/uiData/actions";
+import FieldTypeMultiSelect from "@/components/form/final-form/FieldTypeMultiSelect";
+import type { Option } from "@/components/multi-select/SelectItem";
 
 const VisualizationTypes = {
   MAP: "map",
@@ -123,7 +124,6 @@ const LeaseListPage: React.FC = () => {
   const hasFetchedLeases = useRef(false); // Check if search has been done yet
   const location = useLocation();
   const navigate = useNavigate();
-
   const dispatch = useDispatch();
 
   const isFetching = useSelector(getIsFetching);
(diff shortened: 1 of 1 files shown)


henrinie-nc (commented):
There seems to be a bunch of things going on here, lets try to clarify with questions.

Some extra states are managed here, would it be possible to avoid that completely and keep the state in query parameters instead? It would probably mean something like checking whether service unit filters are set. If on mount they have not been (user didn't use a link with specific filters, or history), then set the default options.
Otherwise use whatever (state) is stored in the queryparams for service unit.

Would that work here, and would we need something more still?

kimi-ihalainen-nc:
@henrinie-nc It was possibly a bit dumb on my part, but I kinda just assumed that the "default" url shouldn't be changed. Meaning if you head to the index page of the app, it wouldn't instantly change the url to something.
If we handle this with url parameters, it will definitely be less complex to implement.

henrinie-nc:
> @henrinie-nc It was possibly a bit dumb on my part, but I kinda just assumed that the "default" url shouldn't be changed. Meaning if you head to the index page of the app, it wouldn't instantly change the url to something. If we handle this with url parameters, it will definitely be less complex to implement.

It might be better to change them to match the actual state anyway, so when user gives that link to someone else, they see the same results

apps/leasing/src/leases/components/LeaseListPage.tsx
```diff
@@ -224,6 +220,27 @@ const LeaseListPage: React.FC = () => {
...
+    if (
+      searchQuery.service_unit === undefined &&
+      userServiceUnits?.length > 0 &&
+      !hasFetchedLeases.current
+    ) {
+      const updatedQuery = {
+        ...searchQuery,
+        service_unit: userServiceUnits.map((unit) => unit.id),
+      };
+      navigate({
```
henrinie-nc, line 237: Could `useSearchParams` and setting new params work here without navigate?

apps/leasing/src/leases/components/LeaseListPage.tsx
```diff
@@ -154,10 +154,10 @@ const LeaseListPage: React.FC = () => {
...
     VisualizationTypes.TABLE,
   );
-  const [serviceUnitOptions, setServiceUnitOptions] = useState<
-    Array<Record<string, any>>
-  >([]);
-  const [selectedServiceUnitOptionValue, setSelectedServiceUnitOptionValue] =
+  const [serviceUnitOptions, setServiceUnitOptions] = useState<Array<Option>>(
+    [],
+  );
+  const [selectedServiceUnitOptionValues, setSelectedServiceUnitOptionValues] =
```
henrinie-nc, line 160: Is this state in the queryparameters and can it be utilized from there without additional state?

apps/leasing/src/leases/components/LeaseListPage.tsx
```diff
@@ -311,10 +328,14 @@ const LeaseListPage: React.FC = () => {
             ? [searchQuery.tenantcontact_type]
             : [];
 
-        if (initialValues.service_unit === undefined) {
-          initialValues.service_unit = "";
-        } else {
-          setSelectedServiceUnitOptionValue(initialValues.service_unit);
+        if (searchQuery.service_unit) {
+          setSelectedServiceUnitOptionValues(
```
henrinie-nc, line 330: These nested ternaries don't look nice, I commented on top of the file about this state setter, is it needed?

apps/leasing/src/leases/components/LeaseListPage.tsx
kimi-ihalainen-nc: The state likely can't be removed easily, as it's used to store the filter options before they're applied to the search. At least debouncing wouldn't work without this.

apps/leasing/src/leases/components/LeaseListPage.tsx
kimi-ihalainen-nc: Even sonarcloud warns about them, so I will fix them

apps/leasing/src/leases/components/LeaseListPage.tsx
kimi-ihalainen-nc: Yes, that is a cleaner approach.

henrinie-nc (approved):
lets go with this!

PR #770, opened 2026-07-07
Add rent share splitting field to new invoice form

kimi-ihalainen-nc:


apps/leasing/src/leases/helpers.ts
```diff
@@ -357,6 +357,33 @@ export const getContentIntendedUse = (
...
+        seen.add(intendedUse.id);
+        return true;
+      }) || [];
+
+  return intendedUses.map((use) => ({
+    value: use.id,
+    label: use.name,
+  }));
+};
+
```
kimi-ihalainen-nc: This one is a bit complex, as it goes through all the tenants and their rent shares, and filters out the duplicates. As in theory, it is possible for all tenants to not share all intented uses.

henrinie-nc: The function name is could describe more of what this is doing, perhaps mention keyword `unique` or similar in its name?

henrinie-nc (approved):
See my comment about naming `getTenantIntendedUseOptions`

henrinie-nc (changes requested):
After testing this in action I realized that it is too vague of a feature. There is no hint that this exists. At least the field should be always visible, but disabled if the feature is not enabled. Then it should explain what it does, and that is quite a long stretch in this implementation.
I could not find a lease in my db that would enable me even testing this, so it seems like quite the edge case:
Is it worth adding this complex logic for an edge case like this? I think we should explore other options of implementing this

kimi-ihalainen-nc:
Based on user feedback and examples, it seems that in practice there should never be a situation where it is important to be able to select a specific intended use manually, so selecting between rent and tenant share should be enough in practically all cases. In some special cases, the invoices can be edited manually.
The new field only shows up if there's actually a possible reason to use it. Meaning there has to be more than one tenant, and tenant shares must exist.
Based on user comments, the feature will be used by ~2 users, so this can be manually explained to them if needed.

apps/leasing/src/leases/components/leaseSections/invoice/forms/NewInvoiceForm.tsx
```diff
@@ -56,11 +61,13 @@ import Loader from "@/components/loader/Loader";
 type InvoiceRowsProps = {
   fields: any;
   useLeaseCreateChargeEndpoint: boolean;
+  isShareableByTenantShare: boolean;
 };
 
 const InvoiceRows = ({
   fields,
   useLeaseCreateChargeEndpoint,
+  isShareableByTenantShare,
```
juho-kettunen-nc: "Jaettava" --> `Divisible` (a whole can be split/divided), rather than `Shareable` (a thing can be shared 
by multiple parties)
EDIT: `Splittable` is better.

juho-kettunen-nc (approved):
Approved because it aligns with latest customer discussions. Just rename the "shareable" to "splittable" to be more understandable.

PR #774, opened 2026-07-10
implements intended uses select grouped by service unit

henrinie-nc:
get intendeduse options directly and group them for Select in lease search

### DIFF
--- a/apps/leasing/src/util/helpers.tsx
+++ b/apps/leasing/src/util/helpers.tsx
@@ -15,6 +15,7 @@ import type {
   Attributes,
   Methods,
   SelectListOption,
+  SelectOptionHds,
 } from "@/types";
 import type { UsersPermissions } from "@/usersPermissions/types";
 import { isBoolean } from "lodash";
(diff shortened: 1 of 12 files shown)


apps/leasing/src/util/helpers.tsx
```diff
@@ -924,3 +925,8 @@ export const debounce = (func: (...args: any[]) => void, wait: number) => {
     timeout = setTimeout(() => func(...args), wait);
   };
 };
+
+export const toHdsOption = (option: SelectListOption): SelectOptionHds => ({
+  label: option.label,
+  value: String(option.value),
+});
```
kimi-ihalainen-nc, line 932: I guess this is an okay solution for now

kimi-ihalainen-nc (approved):

apps/leasing/src/util/helpers.tsx
henrinie-nc: Kinda just to solve a possible type difference as value could be a number in the other case, at least according to the type

PR #776, opened 2026-07-13
Mvj 903 search preparer filter

kimi-ihalainen-nc:
Adds new preparer filter to lease list basic search.
Combines own leases checkbox into an option on the filter


kimi-ihalainen-nc:
This field is currently implemented directly into LeaseListPage. But at some point, this should be converted into using AsyncSelect to allow caching, limited fetching etc. But AsyncSelect and it's related components are currently using class based react & old field components, so those would need to be modernized first.

PR #777, opened 2026-07-14
Fix stale values in LeasePage

kimi-ihalainen-nc:
Also fixes lease area form session storage saving

For some reason a bunch of issues relating to dirty states popped up again. Adding the required values to the refs helped to fix almost all of these. These issues caused the dirty states to display incorrectly and cause session storage functionality to occasionally not store any values.


kimi-ihalainen-nc:
Summertime. Will also merge this without a proper review. Shouldn't contain any changes with possible side effects.

PR #779, opened 2026-07-17
Add preparation state filter

kimi-ihalainen-nc:


kimi-ihalainen-nc:
Currently the replicas of the enum choices are hardcoded into the UI, like some similar fields do, but it would be really nice to figure out a solution to get the choices from the API.

juho-kettunen-nc (approved):

PR #797, opened 2026-08-05
Update tenant list page with new filters and HDS components

juho-kettunen-nc:
New search filters:
* Vuokraustunnus: partial matches from the contact's tenants' leases' identifiers.
* Vain vuokralaiset: only contacts that are a tenant in some lease. Billing contacts are not included.
* Vain aktiiviset asiakkaat: only active contacts in active leases.

Includes modernization to functional components, with HDS styles.
Layout and style of tenant list page now aligns with lease list page.


kimi-ihalainen-nc (approved):
Looks good. Much cleaner than leaselistpage itself, but that one is a lot more complex too.

